PR #10, opened 2013-10-23
Add Sync Value Convenience Function to Sync a Value Between Atoms

blink1073:
I use sync_traits to enforce a relationship between two objects that otherwise do not know about each other.  This behavior could be accomplished with a parent object that monitors the values on both objects and acts as a proxy for the values, but this seems more elegant.  Instead of polluting the Atom namespace, it is used as a convenience function.  I kept the argspec the same as HasTraits.sync_traits.

Here is some example code:

``` python
from atom.api import Atom, Unicode, sync_value


class Dog(Atom):
    first_name = Unicode()

    last_name = Unicode()

    home = Unicode()


class Person(Atom):
    first_name = Unicode()

    last_name = Unicode()

    address = Unicode()


bob = Person(first_name='Robert', last_name='Smith')
dog = Dog(first_name='Fido')

sync_value(bob, 'last_name', dog)
assert dog.last_name == 'Smith'
dog.last_name = 'Richards'
assert bob.last_name == 'Richards'

sync_value(bob, 'last_name', dog, remove=True)
bob.last_name = 'Smith'
assert dog.last_name == 'Richards'

sync_value(bob, 'address', dog, alias='home')
bob.address = '101 Main'
assert dog.home == '101 Main'

sync_value(bob, 'last_name', dog, mutual=False)
dog.last_name = 'Jackson'
assert bob.last_name == 'Smith'
bob.last_name = 'Peters'
assert dog.last_name == 'Peters'
```


blink1073:
Closing as stale.

PR #20, opened 2013-10-26
Add Array Member Type

blink1073:
That was quite fun, thanks for putting up with me.

``` python
from atom.api import Array, Atom
import numpy as np

class Test(Atom):
    arr1 = Array()
    arr2 = Array((1,2,3))
    arr3 = Array([1,2,3])
    arr4 = Array([1,2,3], kwargs=dict(dtype=np.float64))

t = Test()
assert t.arr1.size == 0
assert t.arr2.size == 3
assert t.arr3.dtype == np.int
assert t.arr4.dtype == np.float64 
```

### DIFF
--- a/atom/api.py
+++ b/atom/api.py
@@ -5,7 +5,7 @@
 #
 # The full license is in the file COPYING.txt, distributed with this software.
 #------------------------------------------------------------------------------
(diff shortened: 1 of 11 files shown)


blink1073:
Closing as stale.

PR #25, opened 2013-10-27
Add Alias Class for Delegating Setting/Getting to other Atom objects

blink1073:
Implements #11 using the descriptor protocol.  I needed to add a hook in AtomMeta to set the `attr_name` of the alias, since I could not get it at instantiation time.

``` python
from atom.api import Atom, Instance, Unicode, Alias


class Company(Atom):

    address = Unicode('101 Main')


class Employee(Atom):

    name = Unicode()

    employer = Instance(Company, ())

    work_address = Alias(employer, 'address')


bob = Employee(name='Bob Smith')
assert bob.work_address == '101 Main'
bob.work_address = '23 Pine'
assert bob.employer.address == '23 Pine'
```

### DIFF
--- a/atom/alias.py
+++ b/atom/alias.py
@@ -1,0 +1,31 @@
+#------------------------------------------------------------------------------
+# Copyright (c) 2013, Nucleic Development Team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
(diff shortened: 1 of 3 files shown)


blink1073:
Closing as stale.

PR #63, opened 2015-12-14
Fix #29 Add message filtering to observers

### LINKED ISSUE #29: Add message filtering to observers
It would be useful to filter out messages before calling observers. This can either be a new decorator or an argument to the observe decorator/function:

```
@observe(...)
@filter_on('update')
def _change_handler(self, change):
    ....
```

or

```
obj.observe('...', handler, filter_on='update') 
```

blakev:
Adds the ability to filter `@observe` functions with their change type as well as new-value validation functions.

`@filter_on('create', lambda x: len(x) == 5)` would call the function when the change status is `create` and the new value is five characters long.

### DIFF
--- a/atom/api.py
+++ b/atom/api.py
@@ -5,7 +5,7 @@
 #
 # The full license is in the file COPYING.txt, distributed with this software.
 #------------------------------------------------------------------------------
-from .atom import AtomMeta, Atom, observe, set_default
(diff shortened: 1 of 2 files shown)


MatthieuDartiailh:
This has been addressed for the type change in #155

PR #64, opened 2016-01-25
Fix PyTypeObject instantiations on python 3.5

blakejohnson:
The field definitions of `PyTypeObject` have changed in python 3.5. This adds the necessary `#if/else` statements to populate the object fields appropriately.


sccolbert:
Cheers!

PR #65, opened 2016-03-13
Add support for Python 3 (3.3, 3.4, 3.5)

MatthieuDartiailh:
This makes atom compatible with Python 3 in a python 2 compatible way : 
- it adds future as a dependency (needed for metaclass support, int types, ...)
- I added __future__ imports to all files which may not always be relevant I can remove those if you want. 
- @peterazmanov left some comments in atomlist, I can address them given some pointers.
- Python version checks in C code uses PY_MAJOR_VERSION and PY_MINOR_VERSION rather than the hex representation. @jwiggins used the hex in kiwi. I can move to the hex representation if necessary.
- all iteritems have been removed in favor of items, if performance is a concern for Python 2, future provides a cross-compatible shim that could be used instead.


sccolbert:
This will take some time for me to go through and comment on. I will leave
a note indicating when I'm done commenting. Thanks!

MatthieuDartiailh:
No problem. I will try to make the enaml PRs more manageable. Note that the kiwi PR (https://github.com/nucleic/kiwi/pull/22) is ready too.

MatthieuDartiailh:
Any progress in the review @sccolbert ?

jwiggins:
IIRC, I used the hex version to avoid this particular situation:
```
#if PY_MAJOR_VERSION >= 3
#if PY_MINOR_VERSION > 4
    ( PyAsyncMethods* )0,                  /* tp_as_async */
#else
    ( void* ) 0,                           /* tp_reserved */
#endif
#else
...
#endif
```
I believe my initial reasoning what that it looked a bit crowded, but looking at it again I see another problem. If Python's major version is ever incremented, the minor version will no longer be greater than 4, and then you'll switch back to this being a reserved slot. Probably OK, but certainly not your intent.

MatthieuDartiailh:
Makes sense @jwiggins. I found the hex not super readable but your argument convinced me it is indeed the right thing to do. I will update the code accordingly.

MatthieuDartiailh:
The object declaration now use the PY_VERSION_HEX as suggested by @jwiggins

MatthieuDartiailh:
ping @sccolbert

sccolbert:
Thanks. This is all very appreciated!!! I'm just slammed with other work atm. Sorry :(

blink1073:
Chris is just lazy, don't listen to him /s.

MatthieuDartiailh:
@sccolbert any hope that you will be able to give it a first review soon ?

MatthieuDartiailh:
@sccolbert any hope you could give this a quick look ?

MatthieuDartiailh:
Sorry for bothering you again, but do you have time to give this a look ?

MatthieuDartiailh:
The last commit fix the long standing bug seen in enaml on Python 3 https://github.com/nucleic/enaml/issues/216

atom/api.py
```diff
@@ -1,10 +1,13 @@
 #------------------------------------------------------------------------------
-# Copyright (c) 2013, Nucleic Development Team.
+# Copyright (c) 2013-2017, Nucleic Development Team.
 #
 # Distributed under the terms of the Modified BSD License.
 #
 # The full license is in the file COPYING.txt, distributed with this software.
 #------------------------------------------------------------------------------
+from __future__ import (division, unicode_literals, print_function,
+                        absolute_import)
```
jwiggins: I noticed that you have this in every module in the `atom` package. However I don't think it's needed:
* nothing is using the divide operator
* nothing is printing (examples are another story...)
* `absolute_import`: optional after Python 2.5.0a1, and this project already uses absolute imports
* `unicode_literals`: optional after Python 2.6.0a2, only used in `scalars.py`

See: https://docs.python.org/2/library/__future__.html

I'm not sure what versions of Python should be supported by atom, but it might be OK to not worry about 2.5 & 2.6 support.

atom/scalars.py
```diff
@@ -176,21 +178,35 @@ def __init__(self, default=0.0, factory=None, strict=False):
             self.set_validate_mode(Validate.FloatPromote, None)
 
 
-class Str(Value):
-    """ A value of type `str`.
-
-    By default, unicode strings will be promoted to plain strings. Pass
-    strict=True to the constructor to enable strict string checking.
+class Bytes(Value):
+    """ A value of type `bytes`.
```
jwiggins: You should really comment on the promotion rules here since they also affect this type.

atom/scalars.py
```diff
@@ -202,7 +218,7 @@ class Unicode(Value):
     """
     __slots__ = ()
 
-    def __init__(self, default=u'', factory=None, strict=False):
+    def __init__(self, default='', factory=None, strict=False):
```
jwiggins: I would leave the `u''`

atom/atom.py
```diff
@@ -38,7 +47,7 @@ def observe(*names):
         names = names[0]
     pairs = []
     for name in names:
-        if type(name) is not str:
+        if not isinstance(name, basestring):
```
jwiggins: This also lets bytestrings through. Is that your intent?

atom/atom.py
```diff
@@ -194,7 +203,7 @@ def __new__(meta, name, bases, dct):
         post_validates = []         # Post validate methods: _post_validate_*
         seen_sentinels = set()      # The set of seen sentinels
         seen_decorated = set()      # The set of seen @observe decorators
-        for key, value in dct.iteritems():
+        for key, value in dct.items():
```
jwiggins: I know you mentioned this in the intro for this PR, but I thought I would ask again: If the original version was using `iteritems`, don't you think it's a good idea to preserve that behavior?

I intend for this comment to cover all instances of `iteritems`->`items` and `itervalues`->`values`, but I'll spare you the repeated callouts.

atom/coerced.py
```diff
@@ -17,7 +19,8 @@ class Coerced(Member):
     """
     __slots__ = ()
 
-    def __init__(self, kind, args=None, kwargs=None, factory=None, coercer=None):
+    def __init__(self, kind, args=None, kwargs=None, factory=None,
+                 coercer=None):
```
jwiggins: Technically, nothing _needed_ to be changed in this module

atom/intenum.py
```diff
@@ -1,13 +1,23 @@
...
-import copy_reg
+from __future__ import (division, print_function, absolute_import)
+
+import sys
+if sys.version_info >= (3,):
+    import copyreg
+else:
+    import copy_reg as copyreg
+
+from future.builtins import int as il
```
jwiggins: This name is a bit too terse

atom/src/pythonhelpers.h
```diff
@@ -719,7 +743,7 @@ class PyMethodPtr : public PyObjectPtr {
 
     PyObjectPtr get_class() const
     {
-        return PyObjectPtr( PythonHelpers::xnewref( PyMethod_GET_CLASS( m_pyobj ) ) );
+        return PyObjectPtr( PythonHelpers::xnewref( (PyObject *)Py_TYPE( PyMethod_GET_SELF( m_pyobj ) ) ) );
```
jwiggins: Shouldn't this remain `PyMethod_GET_CLASS( m_pyobj )`?

atom/src/sortedmap.cpp
```diff
@@ -278,17 +277,26 @@ SortedMap_new( PyTypeObject* type, PyObject* args, PyObject* kwargs )
...
+        return 0;
+    }
+#else
+    static void
+    SortedMap_clear( SortedMap* self )
+    {
+        // Clearing the vector may cause arbitrary side effects on item
+        // decref, including calls into methods which mutate the vector.
+        // To avoid segfaults, first make the vector empty, then let the
+        // destructors run for the old items.
```
jwiggins: Maybe move this comment up above the `#if` since it applies to both implementations?

atom/src/validatebehavior.cpp
```diff
@@ -244,10 +244,12 @@ long_handler( Member* member, CAtom* atom, PyObject* oldvalue, PyObject* newvalu
...
 {
-    if( PyLong_Check( newvalue ) )
+    if( Py23Int_Check( newvalue ) )
         return newref( newvalue );
-    if( PyInt_Check( newvalue ) )
-        return PyLong_FromLong( PyInt_AS_LONG( newvalue ) );
+    #if PY_MAJOR_VERSION < 3
+        if( PyInt_Check( newvalue ) )
+            return PyInt_FromLong( PyInt_AS_LONG( newvalue ) );
+    #endif
```
jwiggins: You're leaving this as Python 2 only, but you appear to be changing the behavior.

atom/src/validatebehavior.cpp
```diff
@@ -266,8 +268,10 @@ float_promote_handler( Member* member, CAtom* atom, PyObject* oldvalue, PyObject
 {
     if( PyFloat_Check( newvalue ) )
         return newref( newvalue );
+    #if PY_MAJOR_VERSION < 3
     if( PyInt_Check( newvalue ) )
         return PyFloat_FromDouble( static_cast<double>( PyInt_AS_LONG( newvalue ) ) );
+    #endif
```
jwiggins: Why is this being removed from Python 3?

atom/src/validatebehavior.cpp
```diff
@@ -278,33 +282,58 @@ float_promote_handler( Member* member, CAtom* atom, PyObject* oldvalue, PyObject
...
 }
 
-
 static PyObject*
 unicode_handler( Member* member, CAtom* atom, PyObject* oldvalue, PyObject* newvalue )
 {
     if( PyUnicode_Check( newvalue ) )
         return newref( newvalue );
-    return validate_type_fail( member, atom, newvalue, "unicode" );
+    return validate_type_fail( member, atom, newvalue, "str" );
```
jwiggins: Why the change from `"unicode"` to `"str"`? (also below)

atom/src/validatebehavior.cpp
```diff
@@ -278,33 +282,58 @@ float_promote_handler( Member* member, CAtom* atom, PyObject* oldvalue, PyObject
...
         return newref( newvalue );
+#if PY_MAJOR_VERSION >= 3
+    if( PyBytes_Check( newvalue ) )
+        return PyUnicode_FromString( PyBytes_AS_STRING( newvalue ) );
+
+#else
     if( PyUnicode_Check( newvalue ) )
         return PyUnicode_AsUTF8String( newvalue );
+
+#endif
```
jwiggins: The promotion rules for string types should be clarified.

atom/version.py
```diff
@@ -22,7 +24,7 @@
 
 # The micro release number. The micro release number is incremented
 # for bug fix releases and small feature additions.
-MICRO = 10
+MICRO = 11
```
jwiggins: You think maybe a minor version bump might be in order?

MatthieuDartiailh:
Thanks for the review @jwiggins. I am in the peogress of cleaning this up and your comments are very valuable. I plan to add more tests and enable Travis either in this PR or right after.

sccolbert:
Thanks @jwiggins! This was great feedback.

jwiggins:
I'd highly recommend adding Travis support in a different PR. You've got plenty of changes going on here already.

atom/src/validatebehavior.cpp
MatthieuDartiailh: PyInt_Check does not exist under Python 3 only PyLong_Check

atom/src/validatebehavior.cpp
MatthieuDartiailh: Is not the Python side clear enough ?

MatthieuDartiailh:
I still need to polish a bit. I will add some minimal tests before merging and consolidate the test and add Travis in another one.

atom/src/validatebehavior.cpp
MatthieuDartiailh: Addressed in a version specific manner now. (untetested)

atom/version.py
MatthieuDartiailh: This will be a minor version bump.

atom/atom.py
MatthieuDartiailh: This is debatable. I prefer to have a Python 3 oriented code base. Actually a similar choice was done in Chaco recently.

atom/api.py
MatthieuDartiailh: Usually I prefer to systematically include those imports to be sure nothing breaks unexpectedly. But it add noise to this PR so I removed them as there are not exactly necessary.

atom/src/validatebehavior.cpp
jwiggins: Not quite, no. The docstring for `Str` says:

>Under Python 2 this is a byte string, under Python 3 a unicode one.
>
>The use of this member is discouraged in Python 2/3 compatible codebase
>as Bytes and Unicode provide a more homogeneous behavior.

The other types describe what their promotion semantics are.

The behavior I see here:

INPUT | Unicode | Bytes
----- | ------- | -----
Py2 |  UTF-8 bytestring | Bytes
Py3 | Unicode | Unicode

setup.py
```diff
@@ -62,14 +66,15 @@ def build_extensions(self):
 
 setup(
     name='atom',
-    version='0.3.10',
+    version='0.3.11',
```
jwiggins: `'0.4.0'`

MatthieuDartiailh:
I addressed all comments and ran some basic sanity checks on the examples. I also reverted to using native ints in intenum. I will now work on adding Travis in another PR and merge it first before merging this one. So that we have a proper feedback on it. Compared to the branch I worked for nearly a year the largest change is the handling of integers under Python 3 (so if anybody want to double check, he is welcome).

MatthieuDartiailh:
After quite some tinkering (my Linux VM for python 2 broke so I could not test on Python 2) everything seems to be in order I will merge in a week if nobody complains.

MatthieuDartiailh:
@sccolbert atom is now ready for a new release ! 0.4.0
You just need to update the releasenotes

PR #70, opened 2017-08-10
Add tests

MatthieuDartiailh:
This is still incomplete but aims at covering all Python part of atom and as much of the C part as possible (harder to measure). I will ping when this is ready for review. @sccolbert could you enable Travis on this repo in the mean time I do not have sufficient rights to do it myself.

I chose to group tests by functionalities mostly (changing behaviors, observe, etc) with the exception of the delegator which is a strange beast.


MatthieuDartiailh:
The tests now provide 100% coverage of the Python side (with the exception of _DictProxy that I hope to replace by a C equivalent at some point) and one line in Long which requires a fix on the C side that will be part of the Python 3 support PR. 
I pinpointed what looks like inconsistencies between members (see comments in the tests, mostly test_member) that we may want to address, but it is rather low priority. @sccolbert could you give it a quick look to see if you agree with me ?
In the future I will add coverage for the C++ files and update the tests accordingly.
This is good to go for me. As agreed with @sccolbert I will merge in a week in the absence of comments. Hopefully Travis will be running before that time.

I will clean and rebase the python 3 compatibility PR on top of this work.

One point I noticed when writing the tests is that under Python 3 Range should handle arbitrary integers and hence use PyRichCompare. I do not think it is worth keeping the old behavior for Python 2 so I will use PyRichCompare for both versions (and update the tests).

MatthieuDartiailh:
Continuous integration is running ! As announced I will merge Friday (save if somebody comments of course). Once done I will rebase #65 and merge one week later (if CI is happy and in the absence of comments)

MatthieuDartiailh:
Merging as advertized.

PR #76, opened 2017-11-19
sortedmap: Python 3 fix

MatthieuDartiailh:
Allow to sort any object under Python 3 by sorting by class name and pointer value, which emulates the old behavior on Python 2 more or less.

I have also removed the fallback for sorting object based on direct pointer comparison as it does not make sense from my point of view. I left it for EQ and NE. This does not impact Atom codebase since only Py_EQ was used outside of sortedmap.

I am unsure about the safeguards that should protect the name comparison. It seems to me that an object should always have a type and its type a name but I am opened to suggestions.

### DIFF
--- a/atom/src/pythonhelpers.h
+++ b/atom/src/pythonhelpers.h
@@ -295,15 +295,10 @@ public:
             if( clear_err )
                 PyErr_Clear();
 
-            // FIXME: compare pointers in case of comparison problems
             switch (opid)
             {
-            case Py_LT: return m_pyobj < other;
-            case Py_LE: return m_pyobj <= other;
             case Py_EQ: return m_pyobj == other;
             case Py_NE: return m_pyobj != other;
-            case Py_GT: return m_pyobj > other;
-            case Py_GE: return m_pyobj >= other;
             }
         }
         return false;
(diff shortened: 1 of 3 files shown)


sccolbert:
Comparing based on pointer value is not wrong IMO, once you account for the type names. You have either have to choose an ultimate tie-breaker (pointer value in Python 2), or raise an exception if you can't order two values (Python 3). 

Raising an exception is not an option here, so we must ultimately use pointer values as tie-breakers.

MatthieuDartiailh:
Ok I reworked the comparison fallback could you review my implementation @sccolbert ? I tink we could include a safecmp (or failsafecmp) in cppy mimicking this behavior and always clearing the error (which simplifies a bit the code), WDYT ?

atom/src/pythonhelpers.h
```diff
@@ -24,6 +24,7 @@
 
 #define pyobject_cast( o ) ( reinterpret_cast<PyObject*>( o ) )
 #define pytype_cast( o ) ( reinterpret_cast<PyTypeObject*>( o ) )
+#define pytype_name( o ) ( PyObject_GetAttrString( pyobject_cast( o->ob_type ), "__name__" ) )
```
sccolbert: The getattr call is unnecessary. You can use `o->ob_type->tp_name` directly to get the `char*` type name, and use `strcmp` to compare them.

MatthieuDartiailh: Addressed

sccolbert:
I opened #77 as my proposed fix.

PR #78, opened 2018-03-02
Replace past/future with compat

frmdstryr:
Removes the `future` dependency.

`with_metaclass` and `basestring` were pulled directly from the future and past libraries.

I'm not sure what to think about just using `int` and `long` as I am here. Based on the usage I see in atom and enaml it should be fine but let me know what your thoughts are


MatthieuDartiailh:
I was going to argue that using long was a survivance of Python 2 that I would prefer to avoid but as we still have basestring, this does not really make sense. This is good for me. I will simply change the version number on master and tag 0.4.2.dev and handle the conflicy before merging. Thanks for taking care of this @frmdstryr.

PR #81, opened 2018-07-25
Add testing for Python 3.7

MatthieuDartiailh:


MatthieuDartiailh:
https://github.com/travis-ci/travis-ci/issues/9815

frmdstryr:
Add `dist: xenial` to `.travis.yml`

MatthieuDartiailh:
xenial seems to work in some cases with sudo True, but this means slower builds, so I would prefer to wait a real solution. Otherwise we may consider moving over to CircleCI but I have no experience with it.

frmdstryr:
Yeah, it's still using trusty for some reason...  it's working with [enaml-web](https://travis-ci.org/codelv/enaml-web) but that's not using any sudo setting.

Is using `sudo: True` that much slower?

MatthieuDartiailh:
Sudo is slower yes see https://docs.travis-ci.com/user/reference/overview/. But since this works for enaml-web I will try to insist a bit.

MatthieuDartiailh:
Ok so now we get a real failure on 3.7. And enabling sudo does indeed slow the builds a bit. If we can we should consider getting back to sudo: false at some point in the future.

MatthieuDartiailh:
It turns that due to the use of the fastcall protocol for list.insert/pop/sort in Python 3.7 means that we need to update the C code since we call directly those functions rather than going through PyObject_Call. I am investigating the issue but any help is welcome.

atom/src/atomlist.cpp
```diff
@@ -71,8 +93,14 @@ init_methods()
         py_bad_internal_call( "failed to load list 'reverse' method" );
         return false;
     }
+#if PY_VERSION_HEX >= 0x03070000
+    // typedef _PyCFunctionFast func_t;
```
frmdstryr: Can just the typedef be redefined here here?

Using `typedef _PyCFunctionFastWithKeywords func_t;` builds and passes tests.

frmdstryr: Actually it might be cleaner to define the types the top and use reinterpret cast where needed.
```
#if PY_VERSION_HEX >= 0x03070000
    typedef _PyCFunctionFast func_t;
    typedef _PyCFunctionFastWithKeywords kwfunc_t;
#else
    typedef PyCFunction func_t;
    typedef PyCFunctionWithKeywords kwfunc_t;
#endif
```

atom/src/atomlist.cpp
```diff
@@ -739,7 +779,24 @@ class AtomCListHandler : public AtomListHandler
...
         static char *kwlist[] = { "key", "reverse", 0 };
 #endif
+#if PY_VERSION_HEX >= 0x03070000
+        int nargs = (int)PyTuple_GET_SIZE( args );
+        PyObject **stack = &PyTuple_GET_ITEM( args, 0 );
+
+        PyObject *const *stackbis;
+        PyObject *kwnames;
+        if (_PyStack_UnpackDict(stack, nargs, kwargs, &stackbis, &kwnames) < 0) {
+            return NULL;
```
frmdstryr: Should this return `0`?

atom/src/atomlist.cpp
MatthieuDartiailh: That does seem nicer. I will have a look.

atom/src/atomlist.cpp
MatthieuDartiailh: Both are fine and I simply more or less copy pasted part of CPython C code but for homogeneity in atom 0 may be more appropriate.

MatthieuDartiailh:
Actually I do not like using so much of the private CPython API. 

Looking again at CPython source, we could wrap the MethodDef in a PyCFunctionObject when we create the instance. But that means keeping around many Python objects per instance which is waste of memory. I wish the C API offered a way to call directly a MethodDef without wrapping it first, BU I cannot find one. I will have to keep digging.

frmdstryr:
Could do it how cython does it in https://github.com/cython/cython/blob/master/Cython/Utility/ModuleSetupCode.c#L396-L412

MatthieuDartiailh:
Thanks for the pointer @frmdstryr. But I wish we could avoid diving that deep in the private part of the C API.

MatthieuDartiailh:
I asked on python-list how to avoid the private C API. The answer is to cache the unbound method doing:
```C
list_insert = PyObject_GetAttrString((PyObject *)&PyList_Type, "insert");
...
res = PyObject_CallFunctionObjArgs(list_insert, self, index, value, NULL);
```
This works well for most functions (sort is a bit more painful because it takes kwargs).
I benchmarked both implementations on 3.6:

| Benchmark | py36-opt | py36-generic                        |
| ----------- | ---------- | ---------------------------  |
| insert          | 9.55 us    | 9.85 us: 1.03x slower (+3%)  |
| append       | 88.8 ns    | 149 ns: 1.67x slower (+67%) |
| extend        | 108 ns     | 162 ns: 1.50x slower (+50%)  |

So now the big question is are we willing to pay that king to cost to keep an easier to maintain codebase (and more future proof since fast call is likely to be used in more function in future versions)(knowing that this decision will also impact the dict and set when I finally get there). 
@frmdstryr @sccolbert any opinion ?

MatthieuDartiailh:
@sccolbert I would really like your opinion on this matter, if you have a minute.

MatthieuDartiailh:
ping @sccolbert

MatthieuDartiailh:
In the absence of feedback I will merge later this week and proceed with new releases for atom, kiwisolver and enaml. I plan next to focus on atom for some time to finally incorporate the useful bits from atom-1.0.0 branch.

frmdstryr:
I'm not familiar with how often the private APIs change but my vote is to go with how you currently have it. Less code is better.

Issue #89, opened 2019-02-13
More descriptive errors for Instance members


frmdstryr:
Currently on python 2.7 a TypeError shows a detailed description of what the error is:

```python-traceback
  File "some/path/models.py", line 505, in clone
    clone = Job(**self.members())
TypeError: The 'copy_spacing' member on the 'Job' object must be of type 'list'. Got object of type 'ContainerList' instead.
```

On 3.6 the error is much less useful,

```python-traceback
  File "some/path/models.py", line 505, in clone
    clone = Job(**self.members())
TypeError: invalid instance type
```

Could the same error reporting from 2.7 be used on 3+?

MatthieuDartiailh:
Can you confirm that the difference in the error message message is simply related to the order in which the members are assigned to ? If it is so could you please edit the title of the issue since it seems to be simply related to the `Instance` member and not to the Python version.

frmdstryr:
Oh sorry, changed the title. After adding in #90 it is showing a different member, so the order is changing, good call. 

In either case it would really be helpful to at indicate which member of which member is failing validation for Instance members (and Subclass while we're at it).

frmdstryr:
This is great, thank you!

PR #90, opened 2019-02-13
Use validate_type_fail for instance validation errors

frmdstryr:
For #89 . 

Edit: Should a `validate_instance_fail` be added to have a more clear message?

### DIFF
--- a/atom/src/validatebehavior.cpp
+++ b/atom/src/validatebehavior.cpp
@@ -168,6 +168,24 @@ Member::check_context( Validate::Mode mode, PyObject* context )
 
 
 static PyObject*
+validate_instance_fail( Member* member, CAtom* atom, PyObject* newvalue )
+{
+    PyObject* type = PyObject_Repr( member->validate_context );
+    PyErr_Format(
+        PyExc_TypeError,
+        "The '%s' member on the '%s' object must be an instance of '%s'. "
+        "Got object of type '%s' instead.",
+        Py23Str_AS_STRING( member->name ),
+        pyobject_cast( atom )->ob_type->tp_name,
+        Py23Str_AS_STRING( type ),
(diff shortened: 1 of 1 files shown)


atom/src/validatebehavior.cpp
```diff
@@ -559,7 +559,8 @@ instance_handler( Member* member, CAtom* atom, PyObject* oldvalue, PyObject* new
         return 0;
     if( res == 1 )
         return newref( newvalue );
-    return py_type_fail( "invalid instance type" );
+    PyTypeObject* type = pytype_cast( member->validate_context );
```
MatthieuDartiailh: This is not necessarily correct since an `Instance` member can accept an iterable of types (contrary to a `Typed` member). Having a better error message makes sense but we need a dedicated function for it.

frmdstryr: Oops, thanks! I added one using repr, it's a bit verbose, but I think it should now handle all the difference cases isinstance accepts. I think it'd be acceptable to also just leave it out, just knowing which member of which object is usually enough to trace down the error.

MatthieuDartiailh:
If we start touching this we should do several things:
- update Member::check_context to actually validate the values passed to Instance and Subclass 
- update the error message of Instance. We can only have a type or a tuple of types so for consistency we should not use repr that gives a uselessly long description (but we should use qualname when dropping Python 2)
- update the error message of Subclass.

I will have a bit of time tonight. I will give it a shot.

By the way do you need this on Python 2. I am planning to drop it next time I do any work on Atom (need to finish some work on kiwi first)

frmdstryr:
> By the way do you need this on Python 2.

Not really, the current version works.

frmdstryr:
Closing as #91 does this

PR #83, opened 2018-10-18
Migrate to cppy and limit support to Python 3.5+

MatthieuDartiailh:
Opened mostly to check covergae of c++ using Travis

PR #93, opened 2019-05-08
Typed dict and set in c++

MatthieuDartiailh:
Based on #83


MatthieuDartiailh:
@sccolbert this is ready to merge I believe. Do you want a look first ? 

I would like merge this even though I plan to rework the containers in the future since I don't know when I will have time to do so.

PR #107, opened 2020-07-02
Make all comparisons used in C safe

### LINKED ISSUE #105: Using pandas DataFrame with atom==0.5.1 gives SystemError returned a result with an error set when value changes
After upgrading to `atom==0.5.1` I'm getting `SystemError: ... returned a result with an error set` when any observer is attached to a model or attr that contains a pandas DataFrame in enaml. If I downgrade to 0.4.3 everything works fine (even with the most recent enaml).  

I believe this is because the pandas DataFrame has a custom `__bool__` function that returns an error...  

To reproduce:

```python
import pandas as pd                                                                                                                                                                                                                                                     
from atom.api import Atom, Instance                                                                                                                                                                                                                                     

class Data(Atom): 
    dataframe = Instance(pd.DataFrame)                                                                                                                                                                                                                                                                          

d = Data()                                                                                                                                                                                                                                                              
d.observe('dataframe', print)                                                                                                                                                                                                                                           
d.dataframe = pd.DataFrame()                                                                                                                                                                                                                                            
# {'type': 'create', 'object': <__main__.Data object at 0x7fbb293adcd0>, 'name': 'dataframe', 'value': Empty DataFrame
# Columns: []
# Index: []}

d.dataframe = pd.DataFrame()                                                                                                                                                                                                                                            
# ---------------------------------------------------------------------------
# ValueError                                Traceback (most recent call last)
# ~/miniconda3/envs/test/lib/python3.7/site-packages/pandas/core/generic.py in __nonzero__(self)
#   1478         raise ValueError(
#-> 1479             f"The truth value of a {type(self).__name__} is ambiguous. "
#   1480             "Use a.empty, a.bool(), a.item(), a.any() or a.all()."
#
# ValueError: The truth value of a DataFrame is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().
#
# The above exception was the direct cause of the following exception:
#
# SystemError                               Traceback (most recent call last)
# SystemError: PyEval_EvalFrameEx returned a result with an error set
```

MatthieuDartiailh:
Fixes #105 

We now systematically clear errors occurring during comparison and fallback to 3 way comparison if an error occurred.

@frmdstryr please review the changes and confirm it fixes your issue (I already tested the reproducer from #105).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v1
+        uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
@@ -55,7 +55,6 @@ jobs:
           file: ./coverage.xml
           flags: unittests
           name: codecov-umbrella
-          yml: ./codecov.yml
           fail_ci_if_error: true
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v1
@@ -64,5 +63,4 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
           flags: unittests
           name: codecov-umbrella
-          yml: ./codecov.yml
           fail_ci_if_error: true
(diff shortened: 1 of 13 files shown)


frmdstryr:
Yep working now! Thank you!

MatthieuDartiailh:
Thanks I will try to add some extra tests to cover corner cases in the comparison algorithm before merging. I hope to cut the bugfix release sometimes next week.

PR #122, opened 2021-07-14
Add type hints allowing type checkers to resolve types behind members

MatthieuDartiailh:


MatthieuDartiailh:
@frmdstryr @polysinth I do not know if you have use for static type checking but if you do I would appreciate if you could give this a quick look. I am in particular interested in any suggestion to improve the docs. I am trying to find a balance between giving enough details to be usable but without drowning the user and duplicating all the type hints.

docs/source/basis/typing.rst
```diff
@@ -0,0 +1,184 @@
...
+
+Since a member is a Python descriptor it is logically generic over 2 types:
+
+- the type that will be returned when accessing the member which we will refer to as T
+  in the following
+- the type member can be set to which is not necessarily T since the validation step
+  of a member is allowed to perform generic conversion. We will refer to this type as s
+  in the following.
+
+In the type hints file, the |Member| base class is defined as a generic over T and S.
```
frmdstryr, line 16: At first I was confused as to why there's two types, had to re-read this a few times. 

```suggestion
Since an atom member is a Python descriptor in which the validation step is allowed to perform a type conversion (ex with `Coerced`), the types may be different when reading and writing the member value. Therefore, the type hint is logically generic over 2 types:

- the type that will be returned when accessing the member, which we will refer to as the getter or read type `T`
 
- the type that the member can be set to, which we will refer to as the setter or write type `S`

```

frmdstryr:
I have only recently started adding types so I'm not very familiar yet but I read over the typing doc page you added. I was confused at first but it makes sense after reading a few times. 

This will be great to have, thank you!

frmdstryr:
One somewhat related thing I just hit in a project today (not sure if it is fixable or if I should create an issue)...  but an Atom class can't be used as a mixin to implement an interface defined with the [Protocol](https://docs.python.org/3/library/typing.html#typing.Protocol) type or it gives a metaclass conflict.

sccolbert:
Yes, the Catom authoritative base class cant be used as a mixin. This is a
limitation of the CPython API afaik.

MatthieuDartiailh:
Thanks for having a look @frmdstryr . Your wording seemed indeed a bit more clear so I committed your suggestion.

Regarding your issue regarding multiple inheritance, my first suspect was the existence of AtomMeta since CPython complains about a metaclass conflict. Using `__prepare__` (to set a default value for `__slots__`) and `__init_subclass__` (to handle the members manipulations) introduced since Python 3.6 I believe we could get rid of the metaclass. However there are 2 other blockers, I found trying to make a mixin with Protocol and CAtom or between object and CAtom:
- in the first case it fails with: Protocols can only inherit from other protocols, got <class 'atom.catom.CAtom'>
- in the second with:  Cannot create a consistent method resolution order (MRO) for bases object, CAtom

I still believes eliminating the AtomMeta could be worthwhile but it will not allow what you attempted.

A follow-up I have in mind for this PR is to add a new keyword argument `strict` (possibly keyword only) to Typed, Instance and their forwarded version to disallow None as a value. Currently I am considering that if a default is present the member will never be None but it is not accurate since a user can still assign None  so I think it could be worthwhile to allow for stricter rules. WDYT ?

MatthieuDartiailh:
Following #123 and #124 I am updating the type hints but I have some issues that I would like people opinion on:

- Container typing and member coercion
  Currently one can do : `List(Coerced(str))` meaning that anything supporting string can be added to the list and will be converted on the fly. As a consequence, after adding an element in the list the `in` operator can fail, and similarly remove may not work. The same holds true with even more quirks for dictionary where this mechanism can apply to both key and value. Properly typing this require the members (and the underlying container) to be generic over 2 (4 for dict) type parameters. It is possible but painful. Also I am not sure such a behavior should be encouraged. The other option is simply considering that atomlist[T] = list[T] which is simpler and I think I prefer (but I have a hard time coming to a clear decision). At one point @sccolbert considered disallowing members in containers and I can see some value in that (i.e. allowing validation but not coercion). 

- Validating default values in container members
  For the case of list and sets (assuming 1 generic type parameter), I can get things working in a reasonable number of lines. For Dict the bare typing, is 300 lines long and when I tried previously to type everything (and that approach was simplist and would not have worked) I reached over 1500 lines and mypy becomes quite slow to start. Would it sound reasonable to simply skip checking dict default values (and document it).

an opinion @frmdstryr @clamdad

frmdstryr:
>  The other option is simply considering that atomlist[T] = list[T] 

It'd just keep it simple and do this ^^. 

I'm still not fully versed with python's typing... is there a pattern that can be recommended for more complex cases perhaps that can be suggested? Eg instead of `List(Coerced(str))` recommend creating a custom member and adding typing to that?

> Would it sound reasonable to simply skip checking dict default values (and document it).

I'd just skip it, since atom does validation anyways.

On a side note, I pulled this last night and started tried mypy with atom-db, it seems to be working well so far (picked up a lot of errors) so thank you again for all your efforts!

MatthieuDartiailh:
There is no really simple way to type List(Coerced(str)) since the problem is not only with the member itself but also with the atomlist which has a semantic that is not captured by any Python ABC. At this point I would recommend to people to simply create their own container and stop pretending it is a list. I though about having a member for custom container supporting inner notifications at one point (to circumvent the limitations and over granularity of CantainerList). I will see if I can find my notes and resurrect this.

Creating and typing new members is possible I will add a section to document it. Actually on Python 3.9+, after I add support for generic aliases, it may be actually quite simple. Currently you cannot do it outside a .pyi file which is a bit annoying.

MatthieuDartiailh:
I added support for #128 (not yet merged) and allowed `GenericAliases` to be passed to Typed and Instance members. When doing so the check does not validate the indexed types similarly to how `isinstance(a, list[int])` does not check at runtime that a only contains int.

Apart from some documentation updates, in particular regarding generics and update to the changelog listing in details the backward incompatible changes, this is good to go for me and I would appreciate if somebody can give this a second look.

MatthieuDartiailh:
I updated the docs to reflect support for generic aliases. I will try to merge shortly and then use this on a (closed source) project at work to check I did not miss anything. Please report any issue you encounter with this feature.

frmdstryr:
Thank you. I'm OOTO this week but looked over the test cases and think it looks good. If I find anything ill let you know.

PR #123, opened 2021-08-31
Non optional Typed, Instance and their forwarded versions

MatthieuDartiailh:
This adds a new keyword-only argument to Typed, Instance, and their forward version forbidding the use of None as a value. A new default value mode is also added to provide detailed explanations when trying to access the default value of those member when `optional=False` and no `args`, `kwargs` or factory has been provided.

Note also that the `factory` argument is now keyword-only, I could revert this change but it simplify adding type hints by removing a silly variant.

Finally, all container members use non-optional Instance when wrapping types, since I find that having None implicitly accepted can be rather surprising. Also note that Instance is used over Typed for single type to allow `List(Iterator)` to work properly.

@frmdstryr if you have some time to review this it would be great.

I will update the docs and release notes a bit later.


atom/instance.py
```diff
@@ -15,12 +15,13 @@ class Instance(Member):
     This call is equivalent to `isinstance(value, kind)` and all the
     same rules apply.
 
-    The value of an Instance may be set to None.
+    The value of an Instance may be set to None  if optional is True,
+    otherwise None won't be considered as a valid value.
```
frmdstryr, line 19: `If optional is True, the value of an Instance may be set to None,  otherwise None is not considered as a valid value.`

Maybe?

atom/typed.py
```diff
@@ -14,14 +14,16 @@ class Typed(Member):
     Values will be tested using the `PyObject_TypeCheck` C API call.
     This call is equivalent to `type(obj) in cls.mro()`. It is less
     flexible but faster than Instance. Use Instance when allowing
-    heterogenous values and Typed when the value type is explicit.
+    heterogenous values or (abstract) types relying on custom
+    __isinstancecheck__and Typed when the value type is explicit.
```
frmdstryr, line 18: Need space between `__isinstancecheck__and `

atom/typed.py
```diff
@@ -63,9 +76,9 @@ class ForwardTyped(Typed):
     normal typed.
 
     """
-    __slots__ = ('resolve', 'args', 'kwargs')
+    __slots__ = ('resolve', 'args', 'kwargs', "optional")
```
frmdstryr, line 79: Single quotes to be consistent

frmdstryr:
Looks it fine to me, nice work!

One point to consider, I think it may be more clear to just use `Typed` and `OptionalTyped` (for the cpp values) in the even though it breaks backwards compatibility it's easy to fix and will be more clear going forward.  The `NonOptional` default is fine.

clamdad:
I believe there was an old syntax
`prop = Typed(ClassName, )`
which would force the creation of a `ClassName` instance using the default constructor with no args or kwargs.

Would `optional=False` force the creation of a default object?  If not, what would be the cleanest syntax to achieve that?

frmdstryr:
Including an empty tuple for args will still create a default instance.  So the cleanest would be `Typed(ClassName, (), optional=False)`

atom/instance.py
MatthieuDartiailh: Thanks

atom/typed.py
MatthieuDartiailh: One day I will just `black` everything ... but you are right

MatthieuDartiailh:
@clamdad I agree with @frmdstryr answer to your question.

@frmdstryr The cpp enum is exposed to Python so it would indeed break backward compatibility but the fix would be quite minor and I do not think people used that enum value a lot in the wild so I will do it.

MatthieuDartiailh:
I addressed all the comments and added the docs and release notes. I will merge in the coming days if nobody complains.

frmdstryr:
The forward members clone method needs to copy the optional field. I'm getting an Attribute error in validate. Still need to figure out how to reproduce it in a test

PR #124, opened 2021-08-31
Add strict to FloatRange

frmdstryr:
I've been wanting a FloatRange with strict=False for a while but have just been using Float or manually casting.  

Also unsure if the input parameters be wrapped in float so ints can be passed?

Hopefully I did that right...

### DIFF
--- a/atom/src/validatebehavior.cpp
+++ b/atom/src/validatebehavior.cpp
@@ -702,16 +702,14 @@ PyObject*
 float_range_promote_handler( Member* member, CAtom* atom, PyObject* oldvalue, PyObject* newvalue )
 {
     if( PyFloat_Check( newvalue ) )
-        return float_range_handler(member, atom, oldvalue, newvalue);
+        return float_range_handler( member, atom, oldvalue, newvalue );
     if( PyLong_Check( newvalue ) )
     {
         double value = PyLong_AsDouble( newvalue );
         if( value == -1.0 && PyErr_Occurred() )
             return 0;
-        PyObject* convertedvalue = PyFloat_FromDouble( value );
-        PyObject* result = float_range_handler(member, atom, oldvalue, convertedvalue);
-        cppy::decref(convertedvalue);
-        return result;
+        cppy::ptr convertedvalue( PyFloat_FromDouble( value ) );
+        return float_range_handler( member, atom, oldvalue, convertedvalue.get() );
     }
     return validate_type_fail( member, atom, newvalue, "float" );
 }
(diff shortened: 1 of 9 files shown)


atom/src/validatebehavior.cpp
```diff
@@ -682,6 +682,25 @@ float_range_handler( Member* member, CAtom* atom, PyObject* oldvalue, PyObject*
...
+        return float_range_handler(member, atom, oldvalue, newvalue);
+    if( PyLong_Check( newvalue ) )
+    {
+        double value = PyLong_AsDouble( newvalue );
+        if( value == -1.0 && PyErr_Occurred() )
+            return 0;
+        PyObject* convertedvalue = PyFloat_FromDouble( value );
+        PyObject* result = float_range_handler(member, atom, oldvalue, convertedvalue);
+        cppy::decref(convertedvalue);
+        return result;
```
MatthieuDartiailh, line 698: You can use cppy smart pointer to simplify your life here.

```suggestion
        cppy::ptr convertedvalue( PyFloat_FromDouble( value ) );
        return float_range_handler( member, atom, oldvalue, convertedvalue.get() );
```

atom/src/validatebehavior.cpp
```diff
@@ -682,6 +682,25 @@ float_range_handler( Member* member, CAtom* atom, PyObject* oldvalue, PyObject*
 }
 
 
+PyObject*
+float_range_promote_handler( Member* member, CAtom* atom, PyObject* oldvalue, PyObject* newvalue )
+{
+    if( PyFloat_Check( newvalue ) )
+        return float_range_handler(member, atom, oldvalue, newvalue);
```
MatthieuDartiailh, line 689: spaces
```suggestion
        return float_range_handler( member, atom, oldvalue, newvalue );
```

MatthieuDartiailh (commented):
A couple of suggestions but overall it looks good. Note however that this behavior cannot be statically typed since type checker will always accept an int for a float.

MatthieuDartiailh:
a release notes entry and it will be good to go. I also noticed that `Range` and `FloatRange` are not mentioned in members.rst, if you can fix it great otherwise I will do it.

frmdstryr:
Should I rebase it on top of #123 ? I should have waited

MatthieuDartiailh:
Sure to avoid a conflict in the release notes. Don't worry about the timing, it is nice to see some traffic.

MatthieuDartiailh:
Can you rebase on main now that #123 is in ?

MatthieuDartiailh:
I took the liberty to rebase, I will merge as soon as the CIs come back green.

frmdstryr:
Thank you!

PR #125, opened 2021-08-31
Include optional setting in cloned member and add test

frmdstryr:
Ensure cloned members copy the optional setting.


MatthieuDartiailh:
Thanks for the quick patch !

PR #131, opened 2021-09-22
Make Typed/Instance non-optional by default if a default value can be built

MatthieuDartiailh:
This change is backward incompatible, but the cases in which it can cause issues should be extremely rare. The basic idea is that if somebody write `Typed(A, ())` we will now assumes that None is not a permissible value. My reasoning is that since the value is never None (except in case of an explicit assignment) it is fine to mark the member as non-optional.

Without this, one has to specify both a default construction method and `optional=False` which after experimenting with it feels cumbersome. Any though ?


MatthieuDartiailh:
Any opinion @frmdstryr @clamdad ?

frmdstryr:
Seems reasonable to me.

MatthieuDartiailh:
I will merge next week if nobody raise new concerns. And I will try to cut the release shortly after. I will likely also do an enaml release since the removal of IntEnum broke it and everything will be nicely line with Python 3.10 release.

PR #133, opened 2021-09-30
Generate atom  members from type hints

### LINKED ISSUE #129: Infer members from typehints
Consider inferring members and behavior from python type hint syntax instead of atom.api types.

MatthieuDartiailh:
Closes #129


clamdad:
Wow, I can't wait to try this!  I will definitely take a look this weekend.

MatthieuDartiailh:
@clamdad I fixed the simple Instance case. Feel free to make a PR against this branch with your new examples

MatthieuDartiailh:
I rebased on #136 and the tests are passing again (at least locally).

MatthieuDartiailh:
Mypy is still unhappy about the metaclass though.

MatthieuDartiailh:
@clamdad I believe this is now ready except for expanded documentation and examples. I would appreciate if you could test it.

clamdad:
Consider renaming `tests/typing` module.  Right now, if any tests in `tests` imports `typing`, it will fail to get the build-in python typing package.

clamdad:
Thanks, Matthieu!
See #152 for a new example and test.

MatthieuDartiailh:
@clamdad I added the last safeguards we discussed in #153. I would appreciate a review.

tests/test_atom_from_annotations.py
```diff
@@ -0,0 +1,202 @@
...
+
+    assert not isinstance(A.a, Member)
+
+
+@pytest.mark.skipif(
+    sys.version_info < (3, 9), reason="Subscription of Members requires Python 3.9+"
+)
+def test_ignore_annotated_member():
+    class A(Atom, use_annotations=True):
+        a: List[int] = List(int, default=[1, 2, 3])
```
clamdad, line 69: I'm not sure I understand the intent of this code.  I thought this would be a syntax error.

MatthieuDartiailh: The point is that if the user also defines, for example, a post-validate he may need to narrow down further the type.

What I am trying to avoid is complete garbage annotation like a: dict = Dict() without blocking possibly valid usages.

MatthieuDartiailh: We cannot disallow it in general. Making atom members understandable to type checkers mean you can use a member as a type annotation. One thing I want to allow at least is:

````python

class A(Atom):
 
    o: Member[int, str] = Value()

    def _validate_o(self, old, new):
        assert isinstance(new, str)
        return o

    def _post_getattr_o(self, value):
        return int(value)

```

MatthieuDartiailh:
Setting it to True by default has backward compatibility implications. I have in one of the code base I work with str annotations on some members and having the value to true by default would break it. I agree it is not ideal though. I will try to see if I can fix it easily on my side and revert the need to use the value here.

MatthieuDartiailh:
Checking in my code base, it turns out that if I allow str for member annotations everything is fine. Anyway since we support 3.8 where members cannot have a subscript we have to for consistency. I will push the fix.

MatthieuDartiailh:
Another less common but possible use case is to have a more precise type in annotation for static checking but ignore runtime checking.

tests/test_atom_from_annotations.py
clamdad: Oh, neat.  I didn't know you had implemented that.  My inclination would have been to do something like:

```python
class A(Atom):
    o: Union[int, str]
```

which I would interpret to mean `o = Instance(kind=(int, str))`, and the same validators/getters could be applied if you wanted.

clamdad: Also, I think `Member[int, str] = Value()` has similar issues to combining annotations with member declarations.

MatthieuDartiailh: Actually Member[int, str] means you can set a str but get an int which is different.

MatthieuDartiailh: I find it less problematic than allowing `a: int = Int()` which is always straight up garbage. Here we simply allow user to provide finer grained information to the static type checker that it would otherwise be unable to infer. I find it extremely annoying when atom enforce that typing is correct at runtime but a static type checker fails to understand it. The only case I consider it acceptable is when we are doing value checking in atom.

MatthieuDartiailh: However one thing we can and probably should disallow is a bare member annotation since this is meaningless.

```python
class A(Atom):
    a : Int

```

MatthieuDartiailh: You can have a look at the .pyi for CAtom. Member is defined as a descriptor generic over 2 types (https://github.com/nucleic/atom/blob/main/atom/catom.pyi#L50), the type returned upon access and the type accepted when setting (https://github.com/nucleic/atom/blob/main/atom/catom.pyi#L67 and https://github.com/nucleic/atom/blob/main/atom/catom.pyi#L68).

This is perfectly valid from the point of view of a type checker and does not involve mypy plugin so it should work just as well with pyright.

clamdad: Thanks for pointing that out.  Makes sense now.  Just trying to generalize the usage in my mind.  Why not use the member sublass as the annotation (e.g. `Value[int, str]` or `Instance[int, str]`)?  (I see that instance.pyi declare a single get/set type, but not sure of the underlying restriction).  Or, are you restricting the usage of members as type hints to just the `Member` class.

MatthieuDartiailh:
Currently you cannot specify a default value. 

You earlier mentioned making a default with empty args. I can look into inspecting the type signature to make it viable.

tests/test_atom_from_annotations.py
MatthieuDartiailh: I completely agree. To summarize we have:
```python
# Valid
a: int
a = Int()
a: "Int" = Int()  # -> probably the one we would like to exclude but cannot
a: Int = Int()
# Invalid
a: Int
a: int = Int()
```

clamdad:
I am OK with leaving that feature out for now if you are eager to release.

Otherwise, would this be a viable/simpler alternative to inspection?

```python
class Atom2(Atom):
    y: Atom1 = Atom1(x=5)
```

clamdad:
Agreed.  Let's table it until it's baked.

MatthieuDartiailh:
So anything left blocking this, or should I merge ?

tests/test_atom_from_annotations.py
clamdad: For reference/completeness, atom throws the following exceptions for the others:
```python
a: Int          # ValueError: Member subclasses cannot be used as annotations without specifying a default value for the attribute.
a: int = Int()  # TypeError: Field 'a' of 'A' is assigned a Member-like value but its annotation is not Member compatible
```

clamdad:
If you agree with [this comment](https://github.com/nucleic/atom/pull/133#discussion_r807239719), then I think we're OK.

tests/test_atom_from_annotations.py
clamdad: I've also verified that mypy flags invalid assignments to `a` for the first two cases (e.g. `a.a = "a"` gives the error ` Incompatible types in assignment (expression has type "str", variable has type "int")`

```
a: int
a = Int()
```

MatthieuDartiailh: So the Int stuff is kinda weird. Because atom has a non strict mode accepting float  Int is actually generic over the type that can be set. So my example in addition of being simplistic is actually wrong. I will try to come with something better and add a mypy test.

clamdad:
@MatthieuDartiailh I also converted the observe.py example to a "memberless" version in my branch, which I kinda like, and still serves the purpose:

https://github.com/clamdad/atom/blob/typehint-ex/examples/api/observe_hints.py

This version uses all type hints instead of member declarations. 
MyPy passes strict type checking, except for the `observe` decorator:
```
examples\api\observe_hints.py:32: error: Untyped decorator makes function "any_name_i_want" untyped
examples\api\observe_hints.py:36: error: Untyped decorator makes function "another_random_name" untyped
```

MatthieuDartiailh:
I added the missing type hint to observe and added a test for the valid of:
```python
from atom.api import Atom, List

class A(Atom):
    a: List[int] = List()  # Valid case in which we do not want to pay for runtime validation

```

MatthieuDartiailh:
And turns out the typing around observe is loose... I will look at it when I get the chance.

MatthieuDartiailh:
Everything is fine now . Note that you will have to update your example to use ChangeDict in the annotations of observers.

clamdad:
Thanks for the updates.  This is awesome!

I'll say my final peace on the case below, though.  I see that it's a valid case and achieves static type checking without runtime validation.  But I also see two issues.  First, the penalty for runtime type checking is small (per benchmarks by @frmdstryr).  Perhaps more importantly, as a coding pattern it just doesn't feel right.  It seems very obfuscated unless you know what is going on behind the scenes. 

I'd almost rather see a simple flag passed to the member, like `validate_type=False`, which would be clear and easily documented.

> ```python
> from atom.api import Atom, List
> 
> class A(Atom):
>     a: List[int] = List()  # Valid case in which we do not want to pay for runtime validation
> ```

MatthieuDartiailh:
The penalty of runtime checking can be high for large lists (O(N)).

I agree that we should discourage that kind of usage but I prefer to let it open for advanced user when it is right to use. The use of static type checker (which makes sense if you use annotations) should flag most abuses.

I will merge shortly. Do you think you can add some docs in #153 ?

clamdad:
I finished updating the docs in #153, but that branch was based on this branch.

PR #136, opened 2021-10-13
Support generic aliases in all places where types or tuple of types are allowed

MatthieuDartiailh:
Tests and extra documentation are missing but I otherwise this to be reasonably complete. Feel free to review.

### DIFF
--- a/atom/instance.py
+++ b/atom/instance.py
@@ -5,14 +5,8 @@
 #
 # The full license is in the file LICENSE, distributed with this software.
 #------------------------------------------------------------------------------
-import sys
-from itertools import chain
-
 from .catom import DefaultValue, Member, Validate
 from .typing_utils import extract_types, is_optional
-
-if sys.version_info >= (3, 9):
-    from typing import GenericAlias
 
 
 class Instance(Member):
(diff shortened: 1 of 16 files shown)


MatthieuDartiailh:
@frmdstryr and @clamdad I believe this is finally ready for review and with it 0.7.0 would be ready I believe.

atom/instance.py
```diff
@@ -6,8 +6,10 @@
 # The full license is in the file LICENSE, distributed with this software.
 #------------------------------------------------------------------------------
 import sys
+from itertools import chain
```
frmdstryr, line 9: Unused?

MatthieuDartiailh: Good catch !

atom/instance.py
```diff
@@ -6,8 +6,10 @@
...
 
 from .catom import DefaultValue, Member, Validate
+from .typing_utils import extract_types, is_optional
 
 if sys.version_info >= (3, 9):
     from typing import GenericAlias
```
clamdad, line 15: Also unused?

MatthieuDartiailh:
Thanks for catching those @frmdstryr and @clamdad I will merge and publish a new rc for atom (and merge https://github.com/nucleic/enaml/pull/458 and release a rc for enaml).
I will likely make the final release early next week.

PR #141, opened 2021-11-21
Use linters for the project and modernize the build system

MatthieuDartiailh:
The goal is to use formatters and linters which makes it easier to keep the project style consistent while reducing the burden of doing so. Using more modern build tools also reduce the possible issues with manual version handling while giving more fine grained version during development.


MatthieuDartiailh:
This is in good shape from my point of view. I will merge early next week if I do not hear from anybody by then.

PR #148, opened 2022-01-12
Allow members to be aliased under a different name via a metaclass

frmdstryr:
When a duplicate member (where `a is b`, not a clone) is in the members list, the member index calculation can become invalid as the member is counted multiple times. 

This prevents intentionally "aliased" members from being reindexed.  For context atom-db uses  `_id` as an alias to whatever the member the user tags with `primary_key=True`.  I was getting weird errors in finally traced it down to this...  

Or any please let me know if there's any suggestions on a better way to alias a member...

### DIFF
--- a/atom/atom.py
+++ b/atom/atom.py
@@ -264,12 +264,15 @@ class AtomMeta(type):
         # classes will overlap. When this happens, the members which
         # conflict must be cloned in order to occupy a unique index.
         conflicts = []
-        occupied = set()
+        occupied = {}
         for member in members.values():
-            if member.index in occupied:
-                conflicts.append(member)
+            other = occupied.get(member.index)
+            if other is None:
+                occupied[member.index] = member
+            elif other is member:
+                pass  # Alias to another member
             else:
-                occupied.add(member.index)
+                conflicts.append(member)
 
         # Clone the conflicting members and give them a unique index.
         # Do not blow away an overridden item on the current class.
--- a/tests/test_atom.py
+++ b/tests/test_atom.py
@@ -19,7 +19,7 @@ from textwrap import dedent
 
 import pytest
 
-from atom.api import Atom, Int, MissingMemberWarning, Value, atomref, set_default
+from atom.api import Atom, AtomMeta, Int, MissingMemberWarning, Value, atomref, set_default
 from atom.atom import observe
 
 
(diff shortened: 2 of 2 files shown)


MatthieuDartiailh:
This does not look right to me. We actually have a test ensuring that members are cloned when bound to different names https://github.com/nucleic/atom/blob/main/tests/test_atom.py#L99. What you want is a member transparently forwarding to another which is what Delegator is for.

Also in your case I would suggest modifying the members before running the metaclass rather than after since it looks like a good way to corrupt indexes to me.

So you could do:
```python
class CustomMeta(AtomMeta):
        def __new__(meta, name, bases, dct):
            updated = dict(dct)
            for v in dct.values():
                if isinstance(v, Member) and v.metadata and v.metadata.get("pk"):
                    updated["_id"] = Delegator(v)
            cls = AtomMeta.__new__(meta, name, bases, updated)
            return cls
```

frmdstryr:
Yes that is what I want... thank you!

frmdstryr:
The Delegator seems to squash the original members name?

```py
class A(Atom):
    _id = Int().tag(foo=True)
    id = Delegator(_id)
    

A._id.name
Out[29]: 'id'

A.id.name
Out[30]: 'id'
```

MatthieuDartiailh:
I will look into it, this looks like a bug.

frmdstryr:
I got it to work as is so don't worry about it. 

It would be nice if atom had a way to evaluate an expression before getting or setting (and keeping notifications), while using the same memory, I'm not sure if the Delegator is intended for something like that? 

Eg (this is wrong but you get the idea)

```py
class DiameterAlias(Delegate):
    def getter(self):
         return self.delegate.value * 2
    def setter(self, value):
         self.delegate.value = value/2

class Cylinder(Atom):
    radius = Float(strict=False)
    diameter = DiameterAlias(radius)
```

frmdstryr:
Yes that should work, I thought it still has a copy of the value in that case?  Not that it really matters.  I thought I had tried it a while ago and got weird feedback loops.

MatthieuDartiailh:
This looks like a real bug I will look into it.

MatthieuDartiailh:
Honestly the more I look at Delegator the more broken it feels for your purpose. Typically, there is no way to make the pattern you care about work since when cloned the Delegator will clone the delegate which won't be the clone of the member you care about. So you indeed need to always reassign or create a custom member operating a bit differently.

@sccolbert I would really appreciate some insight in the original purpose of the Delegator member, if you can spare the time.

sccolbert:
I honestly dont fully remember. I think my original intent was to enable it
to delegate to an attribute of a contained (private) object, but i dont
think I ever got around to actually making it able to do that

frmdstryr:
> As a workaround I got it to work by always reassigning the Delegator.

So... this ^^ still does not work... unfortunately I've come full circle and the the index is still getting messed up... 

I had removed the index check when starting to use the Delegator and then hit the original error when testing it in the real application... I re-added the check and its failing here:  https://github.com/codelv/atom-db/runs/4823952918?check_suite_focus=true#step:7:19  

Would it make sense to add an `Alias` member that does this at the catom level?

MatthieuDartiailh:
Interestingly enaml has an Alias member but it does a bit more than just that. Could you check if it would help your case ?

I think it may be worth adding a member doing what you need but we need to design it carefully and determining if we need to extend the working of the metaclass. For example we could allow member to declare whether or not they need storage and only assign them an index if they need one.

Thanks @sccolbert for your answer.

PR #139, opened 2021-11-07
Warn on missing member

MatthieuDartiailh:
Address #137

PR #149, opened 2022-01-13
Include cls in missing member warning

frmdstryr:
Indicate which class has a missing the method. Related to #139


MatthieuDartiailh:
Can you add a changelog entry ?

frmdstryr:
mypy doesn't complain but should the `owner` type be `Type[CAtom]`? I find class vs instance types confusing.

PR #152, opened 2022-01-31
Typehint examples

clamdad:
This branch builds off the [`type-hints` branch](https://github.com/nucleic/atom/tree/type-hints).

It has a simple example and a small test comparing some behavior to dataclasses for basic compatibility.

`tests/typing/test_optional.py`currently highlight two issues:

1. Different behavior when a member/field is declared as Optional without a default (`test_optional_with_no_default`): 
A dataclass will raise a TypeError if the constructor does not have the argument. 
Atom allows it, and sets the default value to None.  See `test_optional_with_no_default` for an example.
Seems like we should decide a preferred behavior and highlight any difference.  I kinda prefer how Atom does it.

2. The default value of Optional members is not respected (`test_optional_with_default_none_value`)
With dataclasses, the default value of an Optional field can include `None`. 
With atom, optional members with a default value raises an error:
`ValueError("Members requiring Instance cannot have a default value.")`
Seems like something we should honor the default value.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,26 +39,25 @@ jobs:
           pip install -r lint_requirements.txt
       - name: Isort
         run: |
-          isort atom -c
-          isort examples -c
-          isort tests -c
+          isort atom examples tests -c
       - name: Black
+        if: always()
         run: |
-          black atom --check
-          black examples --check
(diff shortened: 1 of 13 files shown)


clamdad:
Replaced with #153

PR #153, opened 2022-01-31
Type hint example and test

clamdad:
This branch builds off the [`type-hints` branch](https://github.com/nucleic/atom/tree/type-hints).

It has a simple example and a small test comparing some behavior to dataclasses for basic compatibility.

`tests/typing/test_optional.py`currently highlight two issues:

1. Different behavior when a member/field is declared as Optional without a default (`test_optional_with_no_default`): 
A dataclass will raise a TypeError if the constructor does not have the argument. 
Atom allows it, and sets the default value to None.  See `test_optional_with_no_default` for an example.
Seems like we should decide a preferred behavior and highlight any difference.  I kinda prefer how Atom does it.

2. The default value of Optional members is not respected (`test_optional_with_default_none_value`)
With dataclasses, the default value of an Optional field can include `None`. 
With atom, optional members with a default value raises an error:
`ValueError("Members requiring Instance cannot have a default value.")`
Seems like something we should honor the default value.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# pycharm project files
+.idea
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
--- a/examples/typehints/person.py
+++ b/examples/typehints/person.py
@@ -1,0 +1,51 @@
+# ------------------------------------------------------------------------------
+# Copyright (c) 2013-2022, Nucleic Development Team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# ------------------------------------------------------------------------------
+"""Person tutorial using type hints
+
+"""
+
+from typing import Optional
+
+from atom.api import Atom
+from atom.api import Range
+from atom.api import observe
+
+
+class Person(Atom):
+    """ A simple class representing a person object."""
+
+    last_name: str
+
+    first_name: str
+
+    middle_name: Optional[str] = None
+
+    age = Range(low=0)
+
+    debug: bool = False
+
+    @observe("age")
+    def debug_print(self, change):
+        """ Prints out a debug message whenever the person's age changes.
+        """
+        if self.debug:
+            templ = "{first} {middle} {last} is {age} years old."
+            s = templ.format(
+                first=self.first_name,
+                middle=self.middle_name,
+                last=self.last_name,
+                age=self.age,
+            )
+
+            print(s)
+
+
+if __name__ == '__main__':
+    john = Person(first_name='John', last_name='Doe', age=42)
+    john.debug = True
+    john.age = 43
--- a/tests/typing/test_optional.py
+++ b/tests/typing/test_optional.py
@@ -1,0 +1,78 @@
+from dataclasses import dataclass
+from typing import Optional
+
+import pytest
+
+from atom.api import Atom
+
+
+class InventoryAtom(Atom):
+    """Class for keeping track of an item in inventory."""
+    name: str
+    unit_price: float
+    quantity_sold: Optional[int]
+    quantity_on_hand: int = 2
+
+
+@dataclass
+class InventoryDataclass:
+    """Class for keeping track of an item in inventory."""
+    name: str
+    unit_price: float
+    quantity_sold: Optional[int]
+    quantity_on_hand: int = 2
+
+
+def test_constructor():
+    atm = InventoryDataclass(name='widget1', unit_price=0.99, quantity_on_hand=10, quantity_sold=1)
+    dcl = InventoryAtom(name='widget1', unit_price=0.99, quantity_on_hand=10, quantity_sold=1)
+
+    for obj in [atm, dcl]:
+        assert isinstance(obj.name, str)
+        assert isinstance(obj.unit_price, float)
+        assert isinstance(obj.quantity_on_hand, int)
+        assert isinstance(obj.quantity_sold, int)
+
+
+def test_default_value():
+    dcl = InventoryDataclass(name='widget1', unit_price=0.99, quantity_sold=1)
+    atm = InventoryAtom(name='widget1', unit_price=0.99, quantity_sold=1)
+
+    assert atm.quantity_on_hand == 2
+    assert dcl.quantity_on_hand == 2
+
+
+# @pytest.mark.skip("Optional members with default value should be respected")
+def test_optional_with_default_none_value():
+    """
+    Optional members with default value should be respected
+    """
+
+    @dataclass
+    class TestDataclass:
+        x: Optional[int] = None
+
+    class TestAtom(Atom):
+        x: Optional[int] = None
+
+    dcl = TestDataclass()
+    assert dcl.x is None
+
+    atm = TestAtom()
+    assert atm.x is None
+
+
+# @pytest.mark.skip("Optional members with no default should not be allowed")
+def test_optional_with_no_default():
+    @dataclass
+    class TestDataclass:
+        x: Optional[int]
+
+    class TestAtom(Atom):
+        x: Optional[int]
+
+    with pytest.raises(TypeError):
+        TestDataclass()
+
+    with pytest.raises(TypeError):
+        TestAtom()
--- a/tests/typing/test_optional.py
+++ b/tests/typing/test_optional.py
@@ -42,10 +42,10 @@ def test_default_value():
     assert dcl.quantity_on_hand == 2
 
 
-# @pytest.mark.skip("Optional members with default value should be respected")
 def test_optional_with_default_none_value():
-    """
-    Optional members with default value should be respected
+    """Optional members with default value should be respected
+
+    Fails but seems like it should not.
     """
 
     @dataclass
@@ -62,8 +62,15 @@ def test_optional_with_default_none_value():
     assert atm.x is None
 
 
-# @pytest.mark.skip("Optional members with no default should not be allowed")
 def test_optional_with_no_default():
+    """
+    This example highlights a possible difference between atom and dataclasses.
+
+    When a member/field is declared as Optional without a default:
+    A dataclass will raise a TypeError if the constructor does not have the argument.
+    Atom allows it, and sets the default value to None.
+
+    """
     @dataclass
     class TestDataclass:
         x: Optional[int]
@@ -74,5 +81,5 @@ def test_optional_with_no_default():
     with pytest.raises(TypeError):
         TestDataclass()
 
-    with pytest.raises(TypeError):
-        TestAtom()
+    atm = TestAtom()
+    assert atm.x is None


MatthieuDartiailh:
I agree with your second point and I think it is really easy to fix.

Regarding your first point, it boils down to the fact that atom does not generate a specific. We could add this capability using either a decorator or an extra argument to the metaclass. Either way, we need to think about how we want to handle inheritance.

MatthieuDartiailh:
I addressed your second point in 528e93dc

clamdad:
> I agree with your second point and I think it is really easy to fix.
> 
> Regarding your first point, it boils down to the fact that atom does not generate a specific. We could add this capability using either a decorator or an extra argument to the metaclass. Either way, we need to think about how we want to handle inheritance.

Can you clarify "doesn't generate a specific..."?  

I would hesitate to introduce any new atom-specific user syntax to address issue 1.  Having a "default default" value of None for Optional members seems reasonable, as long as we point it out.

clamdad:
Looks like we have the same Issue 1 above when Optional is not used.  

It this case, it does seem that if a member is *not* optional *and* does not have a default, then we should raise an exception.

```python
def test_no_default():
    @dataclass
    class TestDataclass:
        x: int

    class TestAtom(Atom):
        x: int

    with pytest.raises(TypeError):
        dcl = TestDataclass()

    atm = TestAtom()
    assert atm.x == 0
```

MatthieuDartiailh:
The issue is that subclass of Atom inherit a generic `__init__(self, **kwargs)` and members have same defaults. What you are asking is for the two following cases to behave differently which I dislike, which is why I proposed either a decorator or a new metaclass kwarg to turn on standard dataclass behavior.

```python

# Standard way to build an atom subclass no magic __init__
class TestAtom(Atom):
    x = Int()

# Annotation based atom, here you would like a magic __init__
class TestAtom(Atom):
    x: int

```

clamdad:
Thanks for clarifying - now I understand your concern and what you're proposing.   I agree a "dataclass compatibility" mode makes sense.  A decorator like `from atom.dataclasses import dataclass` would be fine, I think.  Open to other ideas.

Then Atom could also be used in standard mode, where `x: int` behaves like `x = Int()`.  I guess the main difference between dataclass mode and standard mode would be that standard mode adds:
- A default of None for Optional fields
- A "default default" for each field type (e.g. 0 for int or "" for str)

MatthieuDartiailh:
To me, the main difference would be that the dataclass mode generate methods based on members (`__init__`, `__eq__`, etc).

What you call default default always exists and will be what you get for m after doing `del a.m` where a is an Atom subclass instance.

clamdad:
I updated the documentation in the test file in accordance with the above.
Two other things currently worth noting:
1. `test_dataclass_compare/test_list_default` demonstrates a difference regarding mutable defaults.  Is this OK/what we want?
2. `test_dataclass_compare/test_default_list_factory` and `test_default_list_factory/test_default_list_factory_with_default` have some hybrid syntax.  Maybe we want to disallow this?  If we do allow it, the tests should probably pass.

MatthieuDartiailh:
Atom handles properly mutable default since it has copy on assignment semantic for its `List`, `Dict` and `Set` members (necessary to enforce type validation when mutating).  So I would say 1 is what we want.

Regarding 2, the following test do pass for me (and I will add it in #133), do you see failures ? Note however that this tend to confuse type checker and usually requires a `# type: ignore`
```python
def test_ignore_annotated_member():
    class A(Atom):
        a: TList[int] = List(int, default=[1, 2, 3])

    assert A().a == [1, 2, 3]
```

clamdad:
Agreed on mutable defaults! 

Your test passes for me, as does:
```python
def test_default_list_factory_with_default():
    class TestAtom1(Atom):
        x: list[int] = List(int, default=[1, 2, 3])

    with pytest.raises(TypeError):
        TestAtom1(x=[1, 'a', 3])
```
But this fails:
```python
def test_default_list_factory_with_default():
    class TestAtom1(Atom):
        x: list[int] = List(default=[1, 2, 3])

    with pytest.raises(TypeError):
        TestAtom1(x=[1, 'a', 3])
```

MatthieuDartiailh:
Actually this is normal, you do not enforce the type indicated in your annotation in the member. As I said a type checker will flag that and so if you take on you to make such a lie it is on you. Actually you may not want to pay the runtime cost of checking the type but want to allow type checker to pick on error, which is a valid case (usually for more complex content).

BTW this is why container validation stops at depth 1 so `list[list[int]]` will validate that the list elements are list but not that they contain integers. This obviously need to be documented.

clamdad:
Interesting, I see what you are saying.  It's kind of an corner case, but it might surprise/confuse users.  If the reason someone is using Atom instead of dataclasses is for runtime checking, they would expect `x: list[int]` to do runtime checking (as it does), and might not expect `x: list[int] = List(default=[1, 2, 3])` to remove that runtime checking.  Now I can see why it does, but as you said - it would need to be documented, pretty well.

MatthieuDartiailh:
Basically one never needs to write `x: list[int] = List(default=[1, 2, 3])` since `x: list[int] = [1, 2, 3]` just works. And there is no point in annotating a Member assignment except if it needs to contradict what would be inferred from the member.

So I would expect that in practice such a case would be pretty rare, but obviously documented.

clamdad:
Updated test and docstrings to reflect discussion above.

clamdad:
The last thing I can think of to test is inference/behavior of Typed members.
Should the following test fail?
```python
def test_typed():
    """ Example for typed behavior

    """

    class Atom1(Atom):
        x: int

    class Atom2(Atom):
        y: Atom1

    obj = Atom2()
    assert isinstance(obj.y, Atom1)
```

MatthieuDartiailh:
At least I expect it to for now.

My take is that it is dangerous to make a guess at how a default value should be built in such a case. If somebody expect an Iterable I cannot just build one out of thin air. However we could do a bit better by checking for abstract classes and analyzing the signature and if it is fine to call the type with no args (meaning there is no required argument) do that. I will try to implement this.

MatthieuDartiailh:
Also note that since there is no robust way to determine if a class has a custom `__instancecheck__` we always use an `Instance` rather than a `Typed`.

clamdad:
Agree that's dangerous, and we can (and maybe should) use Instance to declare the factory pattern
- `y: Atom1 = Instance(Atom1, kwargs={'x': 2})`

clamdad:
Seems like there may be even more potential for conflicts between the typing declaration and the member declaration.  Could a user argue that the following should not raise the TypeError?

```python
def test_optional_instance_with_constructor_arguments():
    class Atom1(Atom):
        x: int

    class Atom2(Atom):
        y: Atom1

    # Default factory specification
    class Atom2(Atom):
        y: Optional[Atom1] = Instance(Atom1, kwargs={'x': 2})

    assert Atom2().y.x == 2

    with pytest.raises(TypeError):
        Atom2().y = None
```

MatthieuDartiailh:
He could, but the right way to do that is to specify in Instance optional=True.

Once again, a static type checker would complain, which will point out that doing this is not recommended. 

Basically atom is not attrs. One concession we could make would be to introduce special default value marker to allow specifying custom defaults, metadata, and markers for suto generated methods. WDYT ?

clamdad:
I think the static type checker would be happy with the user assigning the value of None (since Optional is in the type spec).  It's Atom that raises the TypeError.  

I thought we might consider having the type spec override the Instance (or other member) argument.  In the previous example, the optional behavior of Instance would be inferred.

MatthieuDartiailh:
The static type checker would complain about assigning a member to the optional. Basically we should discourage standard type annotation and member assignment mixing I think. I don't see any interest in doing it and it opens up just too many issues.

clamdad:
OK, I'm just trying to think of things that folks might try to do and avoid headaches for them. 

Folks coming from dataclasses and wanting to use standard type hints will probably expect a way to declare a factory.  Above I was using the member declaration for that purpose (but I guess that without a special plugin, mypy would declare an type error when assigning a member to that field).  Maybe this is something to consider in the future along with a dataclass decorator.

To keep things simple, it seems better to limit the initial features of atom type-hints.  We can simply declare it illegal to mix type inference with Member declarations.  It might also help to actively raise an exception if a user tries to do it.

MatthieuDartiailh:
We can forbid to use a non-member annotation with a member value (should be fairly straightforward), this should limit the confusion. Out of curiosity how useful is this feature to you without dataclass like features ? I am fine going developing those features in the annotation branch with your feedback.

clamdad:
I think the capability to infer Members from standard Python syntax is very useful.  It makes it easier for new users to build new projects or convert existing projects to leverage the benefits of Atom while minimizing the use of atom-specific member declarations (atom.List, atom.Dict, etc.) throughout their code (which I think is a useful goal for several reasons, including the naming conflict with typing.List and typing.Dict).

That said, I think we should limit those features to only ones that are fully-baked.  So we should definitely prevent the user from using type hints in conjunction with Member assignments until a clear coding/design pattern is chosen.  Handling mutable types is also a nice plus and prevents the need for many factory definitions.

This new coding style (atom with type hints), should be consistent.  For instance, it seems that generally speaking, if the user does not create an ___init__ method, and no kwargs are passed to the constructor, and the type hint does *not* specify a default, the members of a newly constructed object still have a default values of the correct type (or None if Optional).  

The only place this gets tricky is for Instance members (e.g. `x : MyAtom`), which might need a factory if args/kwargs are required.  Currently, we don't have a way to specify that factory using type hints.  And maybe that's also OK.  Maybe we make this fail if the construction of  `MyAtom()` fails. 

With those issues addressed, I think we have a very useful release with less complex features.

For the future,  the idea of a dataclass or similar decorator and the ability to declare factories for instance types would get us really far.  The nice thing is that inheritance should be simpler than it is for dataclasses (don't need to worry about the order of fields with and without defaults) because everything has a default.

clamdad:
@MatthieuDartiailh what would you suggest for this branch?  I am happy to keep it up to date with any changes to your branch, until you think it's finally ready to merge.

MatthieuDartiailh:
I am a bit swamped at the moment, next time I check my branch I will a review of this one. Sorry for the slow progress.

clamdad:
> I am a bit swamped at the moment, next time I check my branch I will a review of this one. Sorry for the slow progress.

Oh my gosh, no rush or need to apologize.  Just hope I'm helping.

MatthieuDartiailh:
I think the new example is nice but need to be updated since use_annotations is not true by default (I have a code base where I triggered some error due to string annotations). The tests however to me are more a comparison to dataclasses than real tests and I would prefer to see such a comparison in the documentation.

examples/typehints/inventory.py
```diff
@@ -0,0 +1,25 @@
...
+
+    name: str
+    unit_price: float
+    quantity_on_hand: int = 0
+
+    def total_cost(self) -> float:
+        return self.unit_price * self.quantity_on_hand
+
+    @observe("unit_price")
+    def _observe_unit_price(self, change):
```
MatthieuDartiailh, line 16: You are requesting to register the observer twice (special name and decorator) also you need a type hint to appease mypy

examples/typehints/inventory.py
```diff
@@ -0,0 +1,25 @@
+from atom.api import Atom
```
MatthieuDartiailh, line 1: Run isort

clamdad:
Moved work to simplified PR: #158

Issue #154, opened 2022-02-10
Proposal: Allow members to selectively set default change event types generated


frmdstryr:
By default atom generates and sends `"created"` event for every member. 

If you profile an app, the [declarative_change_handler](https://github.com/nucleic/enaml/blob/main/enaml/core/declarative_meta.py#L49) causes every widget's `_update_proxy` to be called.  In all default cases this change is discarded leading to lots of function calls/checking that are ultimately ignored.  

A few second benchmark of an enaml-web page generates ~1.5m `_update_proxy` calls (6th most called overall) and which accounts for about ~10% of the init time and could be completely avoided as the proxy is not yet active.

With that said, It'd be nice to have a way to opt in/out of events at the atom level to avoid overhead. Something like:

```python
class A(Atom):
    style = d_(Int(change_types=("update", "delete")))

```

The Member `modes` currently reserves 1 byte for each mode but only the validate mode uses more than 16 values.  

If the other modes were packed into 4 bits there would be room to add a "set_change_types" to selectively enable/disable the default change event types.

Any thoughts? I can work on implementing it.

MatthieuDartiailh:
I have not though about it much but we could also want to filter per observer rather than per member. This way enaml could simply say it does not care about creation event.

frmdstryr:
> filter per observer rather than per member

That makes sense. I suppose a field could need to be added to `ObserverPool::Topic`?

MatthieuDartiailh:
I did not check the code. Feel free to start experimenting in that direction if you want.

frmdstryr:
Changing a single flag gives about a 15% speedup :+1: .

PR #155, opened 2022-02-11
Add observer type filter

frmdstryr:
For #154 . It still needs more testing but any feedback is welcome.

I added this in the declarative_meta 
`member.add_static_observer(declarative_change_handler, 0xF << 1)`

and it avoids about 90% of the `declarative_change_handler` calls (the rest are activated events).  

It is still calling `_update_proxy` the same amount. I think those are all `@observe` handlers so I'm not sure how to make it selectively disable create events on those? Edit: Disabling created events on decorated members it in Atom meta makes it not call `_update_proxy` at all as expected. Any suggestions on how to cleanly tell enaml to do that?


tests/test_observe.py
```diff
@@ -102,6 +102,39 @@ def __call__(self, change):
     assert "str or callable" in excinfo.exconly()
 
 
+@pytest.mark.parametrize(
```
MatthieuDartiailh, line 105: We typically need an IntFlag here, since otherwise it is a mess to read to be honest.

atom/src/observer.h
```diff
@@ -0,0 +1,55 @@
...
+namespace MemberChange
+{
+
+enum Type {
+    Created = 1,
+    Updated = 2,
+    Deleted = 4,
+    Event = 8,
+    Property = 16,
+    Any = 0xFF,
```
MatthieuDartiailh, line 27: I think containers can have another kind of notification too.

MatthieuDartiailh (commented):
I just gave it a quick look but it looks good. You forgot one type of notification for containers (only used by ContainerList). The modes should be exposed on the Python side as a proper IntFlag class. We need to also patch the has_observer(s) methods to take an optional event type for consistency.

frmdstryr:
I'm not sure if I like the name "MemberChange", also should they be named "Create" vs  "Created" etc..?

Should the `observe` also be updated to accept a `change_types` kwarg?

MatthieuDartiailh:
I will think about the names, but it may make sense to stay close to the name used in the change dict (even though I prefer "created" over "create" personally).

Regarding observe, I think we can add a new arg, it should not increase the complexity much. Honestly, I thought a bit about the use case you presented for enaml and my first reaction was that this was the wrong approach and that attaching the observer only when needed would make more sense. However there would be several issues with this approach:
- first it would not handle the case of the declartive_handler
- second it could have a non-negligible memory cost since instead of sharing a single observer for all instances we would have one per instance and given how many observer are used to update the proxy it got me worried.

This is why I think this approach is interesting even though I feel like it could be an interesting foot gun in some situations.

I will try to make a more thorough review of this sometimes next week I hope.

atom/src/observer.h
```diff
@@ -0,0 +1,56 @@
...
+
+#include <cppy/cppy.h>
+#include "utils.h"
+
+
+namespace atom
+{
+
+
+namespace MemberChange
```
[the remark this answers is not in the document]
frmdstryr: Because it imports the fields into the namespace I made a new  `ChangeType` namespace and the type enum fields can be used directly.

MatthieuDartiailh:
Let me know when this is ready for a second round of review.

frmdstryr:
Ready now.

atom/src/member.cpp
```diff
@@ -960,27 +997,43 @@ struct RemoveTask : public BaseTask
...
-    std::vector<cppy::ptr>::iterator it;
-    std::vector<cppy::ptr>::iterator end = static_observers->end();
+    std::vector<Observer>::iterator it;
+    std::vector<Observer>::iterator end = static_observers->end();
     for( it = static_observers->begin(); it != end; ++it )
     {
-        if( *it == obptr || utils::safe_richcompare( it->get(), obptr, Py_EQ ) )
+        if( it->match( obptr ) ) {
+            // TODO: Should it update change types here?
+            it->m_change_types = change_types;
```
MatthieuDartiailh, line 1032: This is a bit tricky. I am inclined to say we should OR the flags, but then the first registered callback may get unexpected notifications and when removing observer we would need to undo the flag. I think the case is rare enough that we could simply error. But I am not sure how to propagate the error...

MatthieuDartiailh (commented):
It looks great but I could only glance quickly. I left one small comment and ideally I would like to see some updates to the documentation and the release notes before merging.

atom/src/member.cpp
frmdstryr: The comment was about whether re-adding should update the change types vs adding another function to do that. I think just updating is the most simple since it checks anyways.

frmdstryr:
What do you think about adding an `__atom_observe_change_types__ = ChangeTypes.ANY` class var to the `Atom` class? 

Enaml's `Declarative` could then update it to exclude create so all the `@observe` decorated handlers ignore create events.

atom/src/member.cpp
MatthieuDartiailh: Previously when adding multiple times the same observer we were simply making we actually connected it a single time. What bothers me here is that code registering the same handler with different change types will race against one another, which is why I would prefer to error if the change types do not match.

MatthieuDartiailh:
I can see the appeal but I need to crawl through the codebase to ensure we are not using static observers for other things.

atom/src/member.cpp
frmdstryr: It also updates the change type when add is called in the observer pool.

MatthieuDartiailh: I don't see your point. The pool is for dynamic observer and I would make the same argument there.

frmdstryr: Was just mentioning it so whatever change needs made is done in both places.  It looks like the insert there was not correct.

frmdstryr: Can you give an example of the problematic case(s) you had in mind?

With how it is now I think there should still only ever be one observer per python object? I tried to keep it that way but maybe I messed something up unintentionally, I don't have much c++ experience.

atom/catom.pyi
```diff
@@ -68,11 +73,19 @@ class Member(Generic[T, S]):
     def __delete__(self, instance: Atom) -> None: ...
     def tag(self: M, **kwargs) -> M: ...
     def clone(self: M) -> M: ...
-    def add_static_observer(self, observer: str | Callable[..., None]) -> Any: ...
+    def add_static_observer(
+        self,
+        observer: str | Callable[..., None],
```
MatthieuDartiailh, line 78: Here we can do better following what I did for observe in atom.py. sorry I forgot to do the same update here.

atom/src/member.cpp
MatthieuDartiailh: One contrived example that popped up in an example written by @clamdad in #158 is:

```python
class InventoryItem(Atom):

    @observe("unit_price")
    def _observe_unit_price(self, change):
        savings = change.get("oldvalue", 0) - change.get("value")
        if savings > 0:
            print(f"Save ${savings} now on {self.name}s!")
```

It is a weird case where one requests an observer on unit_price in 2 ways (through a magic name and a decorator). But if now one writes:

```python
class InventoryItem(Atom):

    @observe("unit_price", ChangeType.CREATE)
    def _observe_unit_price(self, change):
        savings = change.get("oldvalue", 0) - change.get("value")
        if savings > 0:
            print(f"Save ${savings} now on {self.name}s!")
```

now the two declaration are not equivalent and it is far from obvious which one should/will win. This is why I would prefer to simply error in such a case. Since this is the only way add_observer can fail going from void to bool would be sufficient to propagate the error I think. Does that make sense ?

atom/catom.pyi
```diff
@@ -75,14 +75,16 @@ class Member(Generic[T, S]):
     def clone(self: M) -> M: ...
     def add_static_observer(
         self,
-        observer: str | Callable[..., None],
+        observer: str | Callable[[ChangeDict], None],
```
MatthieuDartiailh, line 78: Actually a static observer takes a reference to the owning atom object, see atom.py for an example (tldr you need a TypeVar to work around the function contravariance)

MatthieuDartiailh: Also the fact that the test pass tell me I should improve the type validation tests.

atom/src/member.cpp
MatthieuDartiailh: I did not consider the ModifyGuard I will need to review this more carefully.

atom/catom.pyi
frmdstryr: PyObject_Call must do some magic with the "self" pointer when using a method vs non-method because the test uses
```python
  def react(change):
        changes.append(change)
```

But changing it to this throws an error
```python
  def react(obj, change):
        changes.append(change)
```

MatthieuDartiailh: My bad ! observe can be used on such a callable when it is a method, but at call time it retrieves it by name to get only the bound version.

Sorry for the noise

docs/source/basis/observation.rst
```diff
@@ -63,16 +63,16 @@ Static observers can be bound to a member in three ways:
...
+  which change types to emit
+  + |remove_static_observer| which takes a single callable as argument
 
 Dynamic observers
 ~~~~~~~~~~~~~~~~~
 
 Dynamic observers are managed using the |Atom.observe| and |Atom.unobserve|
 methods of the |Atom| class. To observe one needs to pass the name of the
-member to observe and the callback function. When unobserving, you can either
+member to observe and the callback function.  When unobserving, you can either
```
MatthieuDartiailh, line 75:
```suggestion
member to observe and the callback function. When unobserving, you can either
```

releasenotes.rst
```diff
@@ -24,6 +24,7 @@ Atom Release Notes
    - type_containers: in conjunction with use_annotations determine to what depth
      to type the content of a container. The default is 1 meaning that list[int]
      will use List(Int()) but list[list[int]] will use List(List()).
+- allow specifing which change events are emitted when adding static observers PR#155
```
MatthieuDartiailh, line 27:
```suggestion
- allow specifing which change events are emitted when adding static observers PR #155
```

MatthieuDartiailh (commented):
I have two style nitpicks but otherwise I think we are good. 

People will be able to shoot themselves in the foot if they add twice the same observer with different change types, but given how the observerpool works I don't see how to avoid it and hopefully it won't happen. However you could maybe add a note in the docs saying that adding twice the same observer will overwrite the change type, this way people will be warned.

PR #159, opened 2022-02-18
Fix chained ObserverHandler typing

frmdstryr:
mypy doesn't like chained observers like enaml uses.

### DIFF
--- a/atom/atom.py
+++ b/atom/atom.py
@@ -117,7 +117,11 @@ class ObserveHandler(object):
 
     def __call__(
         self,
-        func: Union[Callable[[ChangeDict], None], Callable[[T, ChangeDict], None]],
+        func: Union[
+            Callable[[ChangeDict], None],
+            Callable[[T, ChangeDict], None],
+            ChangeDict,  # TODO: Workaround for mypy
+        ],
     ) -> "ObserveHandler":
         """Called to decorate the function.
 
--- a/atom/atom.py
+++ b/atom/atom.py
@@ -120,7 +120,11 @@ class ObserveHandler(object):
         func: Union[
             Callable[[ChangeDict], None],
             Callable[[T, ChangeDict], None],
-            ChangeDict,  # TODO: Workaround for mypy
+            # AtomMeta will replace ObserveHandler in the body of an atom 
+            # class allowing to access it for example in a subclass. We lie here by
+            # giving ObserverHandler.__call__ a signature compatible with an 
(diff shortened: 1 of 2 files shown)


examples/api/observe.py
```diff
@@ -37,6 +37,14 @@ def another_random_name(self, change: ChangeDict) -> None:
...
 
 
+class Employee(Person):
+    title = Str()
+
+    @observe("title")
+    def on_title_change(self, change: ChangeDict) -> None:
+        super().any_name_i_want(change)
+
+
```
MatthieuDartiailh, line 47: What is this supposed to demonstrate ?

atom/atom.py
```diff
@@ -117,7 +117,11 @@ def __init__(
 
     def __call__(
         self,
-        func: Union[Callable[[ChangeDict], None], Callable[[T, ChangeDict], None]],
+        func: Union[
```
MatthieuDartiailh, line 120: Looking at the implementation I don't see how it works. Could you point me to an example in the enaml codebase ?

frmdstryr: If you uncomment the "ObserverHandler" then run `mypy examples` it fails when calling the super method.

frmdstryr:
Well this seems like a bug in mypy. It doesn't know how to unroll the observe decorators.

frmdstryr:
https://www.python.org/dev/peps/pep-0612/

atom/atom.py
```diff
@@ -117,7 +117,11 @@ def __init__(
...
+            Callable[[ChangeDict], None],
+            Callable[[T, ChangeDict], None],
+            ChangeDict,  # TODO: Workaround for mypy
```
MatthieuDartiailh, line 123:
```suggestion
            # AtomMeta will replace ObserveHandler in the body of an atom 
            # class allowing to access it for example in a subclass. We lie here by
            # giving ObserverHandler.__call__ a signature compatible with an 
            # observer to mimic this behavior. 
            ChangeDict,
```

MatthieuDartiailh:
I am not sure this really helps us here since the point is that the returned value won't be what is bound to the name later on. 

It iseasier to simply lie here by either saying that the we return the input callable using a TypeVar and use a type: ignore in the return of either observe or observe handler, or lie about the signature of `ObserveHandler.__call__` like you did. The nice part about your lie is that it does not requires a type: ignore. See my suggestion and I think we are good to go.

PR #163, opened 2022-06-03
fully remove the now obsolete setup.cfg

MatthieuDartiailh:


MatthieuDartiailh:
The typin failure exists also on main and will be addressed later so merging.

PR #165, opened 2022-06-08
Fix member mro in the presence of complex multiple inheritance

### LINKED ISSUE #164: Atom has different method resolution order for members
Best explained with a piece of code:
Plain python has the following behavior:

```python

class A:
    a = "a"


class B:
    b = "b"


class AB(A, B):
    pass


class A2(A):
    a = "a2"

class C(AB, A2):
    pass


print(C().a) # => "a2"
```

while atom has the following behavior.

```python
from atom.api import Atom, Str


class A(Atom):
    a = Str("a")


class B(Atom):
    b = Str("b")


class AB(A, B):
    pass

class A2(A):
    a = Str("a2")

class C(AB, A2):
    pass

print(C().a) # => "a"
```

If my understanding is correct, this is because upon cloning the cloned member gets assigned to the class being created https://github.com/nucleic/atom/blob/f0997b31562724e2304ef34e3ab7d5b12cccf9b6/atom/atom.py#L399-L405

Because of this the following lines will also include the cloned members even when they were not defined by the user:
https://github.com/nucleic/atom/blob/f0997b31562724e2304ef34e3ab7d5b12cccf9b6/atom/atom.py#L340-L344

MatthieuDartiailh:
Closes #164 

@jolos since this went unnoticed for so long I would appreciate if you could try to break my fix.

This will need update release notes before merging

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,18 +1,18 @@
 repos:
 - repo: https://github.com/pre-commit/mirrors-isort
-  rev: v5.9.3
+  rev: v5.10.1
   hooks:
     - id: isort
 - repo: https://github.com/psf/black
-  rev: main
+  rev: 22.3.0
   hooks:
     - id: black
 - repo: https://gitlab.com/pycqa/flake8
-  rev: 3.9.2
+  rev: 4.0.1
   hooks:
     - id: flake8
 - repo: https://github.com/pre-commit/mirrors-mypy
-  rev: v0.910
+  rev: v0.960
   hooks:
     - id: mypy
       additional_dependencies: []
--- a/atom/atom.py
+++ b/atom/atom.py
@@ -19,6 +19,7 @@ from typing import (
     Mapping,
     MutableMapping,
     Optional,
+    Set,
     Tuple,
     TypeVar,
     Union,
@@ -258,6 +259,7 @@ class AtomMeta(type):
     """
 
     __atom_members__: Mapping[str, Member]
+    __atom_specific_members__: Mapping[str, Member]
 
     def __new__(  # noqa: C901
         meta,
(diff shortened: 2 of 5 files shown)


jolos:
I'm trying to understand the difference between 'owned_members' and 'specific_members'

https://github.com/nucleic/atom/blob/6c9ff169f0b6d196e58f6012ef373839cae97b6b/atom/atom.py#L350-L353

which sounds similar to 

https://github.com/nucleic/atom/blob/6c9ff169f0b6d196e58f6012ef373839cae97b6b/atom/atom.py#L379-L380

Based on the code, owned_members also includes the inherited members for which the indices conflict?

A more radical approach could be to store the indices not in the members but on the Atom class itself?

MatthieuDartiailh:
This would cause more indirection (while one focus is speed) and may break code relying on the inner working of members so it is not gonna change. 

Otherwise you are right, owned members differs only because it contains cloned member for which there was an index conflict.

jolos:
the `__atom_members__` attribute is now only required for calculating the member count in CAtom_new, so this could be replaced with just a member count attribute. 

The .members method would have to be adapted, but as far as I can tell that should be straightforward.

jolos:
that's true, I was assuming that .members() was less performance critical.  I'm okay with __atom_specific_members__.

Btw note that the also for every getattr/setattr, ... the mro needs to be traversed. It happens in c, so it's plenty of fast. But it's the reason why having long mro's will have a performance impact (at least when using descriptors):

image

MatthieuDartiailh:
Yes but since every member is assigned on the class even if it is not specific to it or owned by it this cost is zero, no ?

I did not recheck the code before writing that.

jolos:
yes, I was kind of expecting that as well. But interestingly enough I get the same performance characteristics with the following:

```python
from functools import reduce
from atom.api import Atom, Str

class Test(Atom):
    a = Str("a2")

def _reduce(Base, _):
    class Test(Base):
        pass
    return Test

Test2 = reduce(_reduce, range(100), Test)

class Test3(Test2):
    a = Str("b2")
```

Which I don't understand to be honest. Though  that probably doesn't have anything to do with atom anyway.

MatthieuDartiailh:
@jolos since this fix your issue I will try to merge shortly. I may however turn __atom_specific_members__ into a tuple of names to avoid duplicating the storage of the members first.

jolos:
> @jolos since this fix your issue I will try to merge shortly. I may however turn **atom_specific_members** into a tuple of names to avoid duplicating the storage of the members first.

Ok, I'll approve! Thx for the quick fix.

MatthieuDartiailh:
I temporarily reverted because I want to cut a patch release for 3.11 but due to the extra work needed for this in #168 I cannot ship it as part of a patch release. I will re-apply the PRs after the release.

PR #182, opened 2022-12-27
Move __getstate__ and __setstate__ to catom

frmdstryr:
If the benchmarks are accurate, this provides a slight improvement (depending on the py version) for serialization speed. 

The current python version can be sped up by moving the slots check into atom meta as done here. If that is done this is still about 15% faster on 3.11.


MatthieuDartiailh:
I do not have time for a complete review (in particular of the C part), but here are just some thoughts.

First, I think it will be very nice to speed up pickling. Second, as we address this there are related two things I would like to address too:
- pickling a frozen object should give back a frozen object
- pickling should not crash on Constant members

Both points should be easy to address but the second may come with some speed penalty. Benchmarks will be handy then.

frmdstryr:
Would it make sense to pass the frozen state as an arg to new https://github.com/nucleic/atom/blob/main/atom/atom.py#L619 ?

I'm not sure if doing it in setstate make senses because the object could be unfrozen (assuming it uses a flag to somehow set the frozen state).

MatthieuDartiailh:
The issue I see is that if we freeze the object in new we won't be able to set any member in setstate. 
I am not sure I follow your argument about freezing based on a flag. If we pickle a frozen atom my naive idea is that the unpickled version should be frozen. So setting the value in setstate does not seem absurd to me.

Out of curiosity did you consider or try to also implement reduce_ex and newargs in C ?

atom/src/catom.cpp
```diff
@@ -370,6 +372,88 @@ CAtom_sizeof( CAtom* self, PyObject* args )
...
+    Py_ssize_t pos = 0;
+    while ( PyDict_Next(membersptr.get(), &pos, &name, &member) ) {
+        cppy::ptr value = selfptr.getattr(name);
+        if (!value || PyDict_SetItem(stateptr.get(), name, value.get()) )
+            return  0;
+    }
+
+    // Frozen state
+    if ( self->is_frozen() && PyDict_SetItem(stateptr.get(), atom_flags, frozen_bit) )
+        return 0;
```
MatthieuDartiailh, line 426: You could do this unconditionally and then pop the key from the state to avoid checking every key. Also we could use a key with a `-` to limit the chances of a conflict with a user member.

atom/src/catom.cpp
```diff
@@ -370,6 +372,88 @@ CAtom_sizeof( CAtom* self, PyObject* args )
...
+    if ( !membersptr || !PyDict_CheckExact( membersptr.get() ) )
+        return cppy::system_error( "atom members" );
+
+    PyObject *name, *member;
+    Py_ssize_t pos = 0;
+    while ( PyDict_Next(membersptr.get(), &pos, &name, &member) ) {
+        cppy::ptr value = selfptr.getattr(name);
+        if (!value || PyDict_SetItem(stateptr.get(), name, value.get()) )
+            return  0;
+    }
```
MatthieuDartiailh, line 422: Here since we have a direct access to the member we could check the setattr mode and exclude member with a constant mode, signal mode, property mode and a None context which we are sure are read only. 
Alternatively we could add an attribute on the member signaling if it can be pickled. This would allow to keep using the C getstate with custom members. WDYT ?

atom/src/catom.cpp
```diff
@@ -370,6 +372,98 @@ CAtom_sizeof( CAtom* self, PyObject* args )
...
+    PyObject *name, *member;
+    Py_ssize_t pos = 0;
+    while ( PyDict_Next(membersptr.get(), &pos, &name, &member) ) {
+        switch (member_cast(member)->get_setattr_mode()) {
+            case SetAttr::Mode::Constant:
+            case SetAttr::Mode::ReadOnly:
+            case SetAttr::Mode::Event:
+            case SetAttr::Mode::Signal:
+            case SetAttr::Mode::Property:
+                continue;
```
MatthieuDartiailh, line 425: Your approach here look a bit too broad to me. ReadOnly can be restored when unpickling but it should be only if a value was actually set. Similarly a Property could be restored if it is writable (but it is more debatable). This really makes me think a custom method of the member could be useful.

atom/src/catom.cpp
```diff
@@ -370,6 +372,98 @@ CAtom_sizeof( CAtom* self, PyObject* args )
...
+    cppy::ptr itemsptr = PyMapping_Items(state);
+    if ( !itemsptr )
+        return 0;
+    cppy::ptr selfptr(pyobject_cast(self), true);
+    bool freeze = false;
+    for ( Py_ssize_t i = 0; i < PyMapping_Size(state); i++ ) {
+        PyObject* item = PyList_GET_ITEM(itemsptr.get(), i);
+        PyObject* key = PyTuple_GET_ITEM(item , 0);
+        PyObject* value = PyTuple_GET_ITEM(item , 1);
+        if ( PyObject_RichCompareBool(key, atom_flags, Py_EQ) == 1 && PyLong_Check(value) )
```
MatthieuDartiailh, line 456: I still believe there could be value in making the frozen key always present and to pop it early on.

frmdstryr: There didn't seem to be any measurable performance impact when adding the extra compare. Maybe it will if theres a lot of fields, idk.

Also setstate can be used outside of pickle (my use case does not involve pickle at all) so there's no guarantee that the flag will be there.  In my case state is not a dict but a row from sqlalchemy that implements a mapping.

MatthieuDartiailh: Fair enough for your use case but could you add a comment about it then ?

frmdstryr: It now just checks if the flag is there (and deletes it)  which should work fine for my use case since that will always be skipped.

Do you think it should put the flag back in so the state is left unchanged?

MatthieuDartiailh: So I tried to look up what invariant setstate is expected to maintain but did not find much. Basically it seems from the docs that when a class defines a custom getstate it can expect to get the getstate output as input of setstate.  

So assuming mutability sounds reasonable but this does not really align with your use case. Could you maybe point me to the origin of the behavior you are relying on ?

MatthieuDartiailh:
For the sake of readability I would use `--frozen` instead of `-f` to mark an instance is frozen.

MatthieuDartiailh:
Below I try to summarize reasons and ways we may want to customize when a member value is made part of an Atom state.

Tuning of getstate behavior:
- by member type:
  - most members: always
  - Constant, Signal: never 
  - ReadOnly: if the member was set previously (`cppy::ptr value( atom->get_slot( member->index ) );`)
  - Property: if settable
  - custom: by defining a custom member
- in a user customizable way:
  - option 1: use a tag. Could conflict with an existing metadata use but is simpler. Simple when the answer is unconditional and involve less overhead.
  - option 2: use a method on the atom object like for any other member customization. More work in particular on the metaclass.
  - do both

No matter what we do for the last bullet point I feel that having a getstate mode like we do for getattr, setattr, validate etc is the way to get enough versatility and won't be surprising to people using Atom. The enum value would be:
- Include
- Exclude
- IncludeNonDefault  (for ReadOnly)
- Property
- MemberMethod_Object  (for user custom members)
- ObjectMethod_Name (for customization from an Atom instance)

The method to retrieve whether or not a member should be included would take the atom instance and return a bool. I think the metadata value should take precedence and if set we should ignore the mode. We cannot easily bypass looking the key in teh dictionary since the metadata dictionary is mutable.

@frmdstryr WDYT of the above ? 

It is definitively more work than what you have already done but I think it would be worth it. I can help with any part of the above plan if you want.

frmdstryr:
This is all kind of more than I intended for with this PR... (just slightly faster version of the current behavior).

Anyways, in my opinion if atom is going to have a way to customize the state I don't think choosing whether the value is included or not is flexible enough. It's fine for pickle but not for serializing to something like json which cannot handle py objects. 

With atom-db I added a way to tag a member with custom functions to serialize/deserialize the member (eg https://github.com/codelv/atom-db/blob/master/tests/test_json.py#L35). IMO that is the best. Its simple, yet still flexible because the user has full control on what goes in and out.

MatthieuDartiailh:
I am sorry I am somehow hijacking your PR but the current behavior has already bitten me several times (in particular the impossibility to pickle an Atom with a Constant member) and I saw this PR as a good opportunity to fix the "broken" parts.

What bothers me about your approach is that `__getstate__` is meant for pickle not to serialize to a different format. I have used a trick similar to yours to serialize atom object to INI but I feel this is orthogonal to `__getstate__` and pickle.

If you need to go through your custom serialization, deserialization function, am I right to think that this PR does not help with speed ?

MatthieuDartiailh:
@frmdstryr could you comment on how a faster getstate in C helps with your case of using tags ? 

For the pure pickle use case, I believe the plan I proposed make sense (except we could even avoid the tag and the issue of using a name other people may already rely on). I will try to work on it when I have the cycles.

frmdstryr:
You're correct the speed difference is negligible. Adding the enum you proposed makes sense.

MatthieuDartiailh:
Could you enable edit by maintainers ? I have implemented the changes described above but cannot push.

frmdstryr:
It is checked? I have no branch protection stuff setup, but maybe its because its the main branch? IDK. 
image

Feel free to just close and open a new one.

MatthieuDartiailh:
my bad I worked on your frmdstryr/move-get-set-state-to-catom branch and failed to notice that the PR was from main...

frmdstryr:
Oh, I had to move it to main to get actions to run.

MatthieuDartiailh:
I will go through the changes again to check I do not mess up something while rebasing/cherry-picking but otherwise this should be good to go (modulo the missing documentation). @frmdstryr can you have a look.

atom/src/catom.cpp
```diff
@@ -370,6 +371,105 @@ CAtom_sizeof( CAtom* self, PyObject* args )
...
+    PyObject *name, *member;
+    Py_ssize_t pos = 0;
+    while ( PyDict_Next(membersptr.get(), &pos, &name, &member) ) {
+        cppy::ptr should_gs = member_cast( member )->should_getstate( self );
+        if ( !should_gs ) {
+            return 0;
+        }
+        int test = PyObject_IsTrue( should_gs.get() );
+        if ( test == 1) {
+            PyObject *value = PyObject_GetAttr( selfptr.get(), name );
```
frmdstryr, line 424: Not sure if this is correct? Shouldn't it  just continue if test == 0 or return if  test== -1?

MatthieuDartiailh: Yes it was a bad rebase, should be fixed now.

atom/src/catom.cpp
```diff
@@ -370,6 +371,106 @@ CAtom_sizeof( CAtom* self, PyObject* args )
...
```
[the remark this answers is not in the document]
MatthieuDartiailh: Yep, I will change it it (all those comes from me working on a non up to date branch sorry).

I will look into the other nasty errors we are seeing in the CIs.

MatthieuDartiailh:
Thanks for having a look @frmdstryr . I still need to go through everything once but I am wondering, should we add support for magic method name in AtomMeta for this use case ? 

It may clash with existing names which would make the change more intrusive than I would like but not having it will be surprising since it would the only behavior missing it. Do you have an opinion ?

frmdstryr:
Agree. I would just add it to the customization page.

MatthieuDartiailh (approved):

MatthieuDartiailh:
Thanks @frmdstryr for initiating the work on this. It is always extremely motivating for me to work on a feature I know you need.

frmdstryr:
Thanks!

PR #200, opened 2023-09-05
Start running tests on 3.12

MatthieuDartiailh:


MatthieuDartiailh:
After investigating locally it appears that the segfault on exit are linked to cleaning up C++ namespaces containing Python object late in the interpreter shutdown process. The proper workaround is likely to store such variables in a module state (as should be done for a bunch of other global variables).

This may require to drop 3.8 and to do some funny dance on 3.9 and 3.10 since we cannot get a module state in a slot in those versions.

MatthieuDartiailh:
I found a workaround for the segfault but we should move to using module states still.

PR #204, opened 2023-10-03
Fix typo in add_member

frmdstryr:
Looks like behavior was accidentally changed from https://github.com/nucleic/enaml/commit/58b65c2df870ed77eb06b28466302228088332d3

Fixes https://github.com/nucleic/enaml/issues/534

### DIFF
--- a/tests/test_atom.py
+++ b/tests/test_atom.py
@@ -164,6 +164,29 @@ def test_add_member():
     assert "a" in B().members()
 
 
+def test_add_member_overridden_member():
+    class A(Atom):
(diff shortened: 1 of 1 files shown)


MatthieuDartiailh (approved):

MatthieuDartiailh:
I added a test for the case of copying static observers. 

Note that this is a weird case since we do not copy other behaviors. I will merge as soon as the CIs are green and I will make a new bugfix release.

frmdstryr:
Thanks again

PR #205, opened 2023-11-20
Use ruff for linting and formatting

MatthieuDartiailh:


MatthieuDartiailh:
Some extra tests should be added since the reformatting of lambdas allowed to identify some untested areas

PR #210, opened 2024-03-18
Apply dataclass transform to AtomMeta

### LINKED ISSUE #173: Investigate support for PEP681 : dataclass_transform
This would help atom integrate better with static type checker.

MatthieuDartiailh:
All members are declared as field specifiers which allow to use them in addition of an annotation.

I am considering adding other field specifiers to:
- allow to tag a member without needing to rely on a member
- allow to specify a default value factory
- I would like to be able to specify arg/kwargs for Typed/Instance but this cannot be statically validated...

I am considering something along the following lines to stay close to the existing syntax.

```
class A(Atom):

    a: int = member(default=1).tag(a=1)
  
```

Any opinion @frmdstryr

Closes #173

### DIFF
--- a/tests/test_atom_from_annotations.py
+++ b/tests/test_atom_from_annotations.py
@@ -309,7 +309,7 @@ def test_annotated_containers_no_default(annotation, member, depth):
         (Dummy, Coerced, member().coerce(lambda v: Dummy(v, 5))),
         (Dummy, ReadOnly, member(1).read_only()),
         (Dummy, Constant, member(1).constant()),
-    ]
+    ],
 )
 def test_annotations_with_default(annotation, member_cls, default):
     class A(Atom, use_annotations=True):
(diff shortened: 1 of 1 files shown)


sccolbert:
I don't think I'm following what you want here. Could you please explain a bit more with the some code examples of what you are trying to achieve?

MatthieuDartiailh:
My goal is overall to make Atom feel closer to other technology such as the stdlib dataclass. A first step that will improve the interaction with IDE/type checker is to use the dataclass_transform on the atom metaclass: https://typing.readthedocs.io/en/latest/spec/dataclasses.html#the-dataclass-transform-decorator

In order to achieve that I want to never use the Member subclasses explicitly and only use type annotations and field specifiers (https://typing.readthedocs.io/en/latest/spec/dataclasses.html#field-specifiers).

sccolbert:
I think this would be a great change for the next version of `Atom` where we can introduce some breaking changes. That would let us cleanup/remove the members that don't make sense, while also simplifying the C++ code base.

For example, we could have just one `Member` class behind the scenes which has an enum defining its type. That type could be type-checked when running in a "debug" mode, but be cost-free when running in production.

sccolbert:
Another thing I want to test is the space/perf efficiency of the new(er) Python dict implementation for instance attributes. I've read that it's quite an improvement over Python 2.x, but I've been wondering how they compare to Atom for space and performance.

MatthieuDartiailh:
> Another thing I want to test is the space/perf efficiency of the new(er) Python dict implementation for instance attributes. I've read that it's quite an improvement over Python 2.x, but I've been wondering how they compare to Atom for space and performance.

Size wise the last I did the comparison an Atom instance was still smaller than a Python class but larger than a slotted dataclass. Speed wise I need to redo some benchmarks.

PR #211, opened 2024-03-18
Add FixedTuple member enforcing a given number of items

MatthieuDartiailh:


MatthieuDartiailh:
@frmdstryr do you have an interest in this feature and do you want to review ?

atom/src/validatebehavior.cpp
```diff
@@ -463,6 +482,56 @@ tuple_handler( Member* member, CAtom* atom, PyObject* oldvalue, PyObject* newval
...
+{
+    if( !PyTuple_Check( newvalue ) )
+    {
+        return validate_type_fail( member, atom, newvalue, "tuple" );
+    }
+    cppy::ptr tupleptr( cppy::incref( newvalue ) );
+
+    // Create a copy in which to store the validated values
+    Py_ssize_t size = PyTuple_GET_SIZE( newvalue );
+    cppy::ptr tuplecopy = PyTuple_New( size );
```
frmdstryr, line 496: This could be moved to after the size check

MatthieuDartiailh:
That's a good point. I will look into it. The error message may indeed be more challenging but I will look into it.

MatthieuDartiailh:
So the issue with re-using the same member is that it is not backward compatible. Changes to how we interpret annotations makes the system stricter but it bothers me less since it should not break anything that was previously valid.

frmdstryr:
I am normally against breaking backwards compatibility but in this case I think it is worth it. I'm so used to using the `tuple[x, y, z]` with typing I actually found an invalid use of the current `Tuple` that passes what should be the second type as the default value. Maybe others can chime in but do whatever.

MatthieuDartiailh:
Thanks @frmdstryr I will keep two separate members. 

Please note that if you use annotations instead of the actual members `tuple[x, y, z]` will use `FixedTuple` which should fix your issue.

MatthieuDartiailh:
I forgot to address your comments (forgot to re-read the whole discussion). I will come back to it once I am done with #210

PR #218, opened 2024-12-29
Accept NewType instances in all spots types are accepted

MatthieuDartiailh:


MatthieuDartiailh:
@sccolbert do you want to have a look or can I proceeed with merging this ?

PR #219, opened 2025-01-07
Allow the use of Literal in type annotation and use Enum in those cases

MatthieuDartiailh:


MatthieuDartiailh:
@sccolbert do you want to have a look or can I proceeed with merging this ?

PR #221, opened 2025-01-09
Change range/float range to value error and include member/obj in error message

frmdstryr:
Currently a value out of range throws a type error even if it is the correct type. 

This also includes the member name in the error message.


frmdstryr:
Well it seems something changed in the float_range_handler with 3.12. 

I started getting 'too small' validation errors with `FloatRange(low=0, strict=False)` for a value of 25. After updating the message to print the low value it was some huge value?

frmdstryr:
If I change low value to a float `low=0.0` it works it seems the low & high types are not checked.

MatthieuDartiailh:
Can you rebase on main to fix the CI issue. Also it may be worth casting the low, high to preserve the fast validation path.

frmdstryr:
The problem was there was no context validation for FloatRangePromote since the case was missing.

MatthieuDartiailh (approved):

PR #223, opened 2025-01-18
Refactor defaultvaluebehavior to avoid unnecessary allocations

frmdstryr:
All of the tuple creations can be eliminated by using either vectorcall or the builtin call functions.

PR #224, opened 2025-01-18
Refactor getattrbehavior to avoid allocations

frmdstryr:
Same as #223 but separate PR to make it easier to review.

### DIFF
--- a/atom/src/getattrbehavior.cpp
+++ b/atom/src/getattrbehavior.cpp
@@ -161,17 +161,7 @@ _mangled_property_handler( Member* member, CAtom* atom )
     cppy::ptr name( PyUnicode_FromFormat( "_get_%s", suffix ) );
     if( !name )
         return 0;
-    cppy::ptr callable( PyObject_GetAttr( pyobject_cast( atom ), name.get() ) );
-    if( !callable )
-    {
-        if( PyErr_ExceptionMatches( PyExc_AttributeError ) )
-            PyErr_SetString( PyExc_AttributeError, "unreadable attribute" );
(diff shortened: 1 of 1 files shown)


frmdstryr:
I think if catom is updated to use the fastcall protocol more optimizations can be done here.

frmdstryr:
Could the mangled name "_get_%s" somehow be cached instead of reformatting it every time?

sccolbert:
> Could the mangled name `_get_%s` somehow be cached instead of reformatting it every time?

Can you link to a line number where you mean?

frmdstryr:
Here https://github.com/nucleic/atom/blob/main/atom/src/getattrbehavior.cpp#L161

Idk if the getattr_context can be repurposed to store the name?

sccolbert:
I don't think that's worth the tradeoff. The `Property` members aren't used very often, and while they could benefit from some more storage for caching, that would complicate the entire `Member` struct, which tries to be as space-efficient as possible.

When I originally wrote `atom`, it was for a commercial purpose sponsored by my employer. In that setting, optimizing memory use was the main driver. They have a very large codebase with tens of millions of Python objects instantiate at any given time. Since they are still using the library AFAIK, I don't want to make changes that would explode memory usage for them.

MatthieuDartiailh (approved):
This is a very nice diff !

sccolbert (approved):

MatthieuDartiailh:
I will restart the CIs later (the API rate limits for external contributor is so annoying) and merge when they come back green.

PR #225, opened 2025-01-18
Refactor delattrbehavior calls

frmdstryr:
Same thing for delattr

### DIFF
--- a/atom/src/delattrbehavior.cpp
+++ b/atom/src/delattrbehavior.cpp
@@ -155,17 +155,7 @@ _mangled_property_handler( Member* member, CAtom* atom )
     cppy::ptr name( PyUnicode_FromFormat( "_del_%s", suffix ) );
     if( !name )
         return -1;
-    cppy::ptr callable( PyObject_GetAttr( pyobject_cast( atom ), name.get() ) );
-    if( !callable )
-    {
-        if( PyErr_ExceptionMatches( PyExc_AttributeError ) )
-            PyErr_SetString( PyExc_AttributeError, "can't delete attribute" );
-        return -1;
-    }
-    cppy::ptr args( PyTuple_New( 0 ) );
-    if( !args )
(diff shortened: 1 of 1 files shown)


sccolbert (approved):

atom/src/delattrbehavior.cpp
```diff
@@ -155,17 +155,7 @@ _mangled_property_handler( Member* member, CAtom* atom )
     cppy::ptr name( PyUnicode_FromFormat( "_del_%s", suffix ) );
     if( !name )
         return -1;
-    cppy::ptr callable( PyObject_GetAttr( pyobject_cast( atom ), name.get() ) );
-    if( !callable )
-    {
-        if( PyErr_ExceptionMatches( PyExc_AttributeError ) )
-            PyErr_SetString( PyExc_AttributeError, "can't delete attribute" );
```
MatthieuDartiailh, line 162: Losing this means we lose on the quality of the error message. It is sad we cannot preserve it while improving performance.
Would creating the method and then calling it with a better PyObject_Call make it much worse than the current solution.

sccolbert:
I thought about this before I approved the changes, but the attribute
exception error I think would be "sufficient enough" since it would be
something like "object has no attribute `_del_foo`".

frmdstryr:
Looks ok to me if I did this right..

```
In [23]: class A(Atom):
    ...:     y = 1
    ...:     def  _get_x(self):
    ...:         return self.y
    ...:     x = Property()
    ...: 

In [24]: A.x.set_delattr_mode(DelAttr.Property, None)

In [25]: a = A()

In [26]: a.x
Out[26]: 1

In [27]: del a.x
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[27], line 1
----> 1 del a.x

AttributeError: 'A' object has no attribute '_del_x'


```

PR #227, opened 2025-01-18
Refactor setattrbehavior calls

frmdstryr:
Please double check my usage of valueptr. It used `valueptr.release()` before but since the args doesn't steal a reference I believe `valueptr.get()` is correct.

### DIFF
--- a/atom/src/setattrbehavior.cpp
+++ b/atom/src/setattrbehavior.cpp
@@ -297,7 +297,8 @@ call_object_object_value_handler( Member* member, CAtom* atom, PyObject* value )
     if( !valueptr )
         return -1;
     PyObject* args[] = { pyobject_cast( atom ), valueptr.get() };
-    if( !PyObject_Vectorcall( member->setattr_context, args, 2, 0 ) )
+    cppy::ptr ok( PyObject_Vectorcall( member->setattr_context, args, 2, 0 ) );
+    if( !ok )
         return -1;
     return 0;
 }
@@ -310,7 +311,8 @@ call_object_object_name_value_handler( Member* member, CAtom* atom, PyObject* va
     if( !valueptr )
         return -1;
     PyObject* args[] = { pyobject_cast( atom ), member->name, valueptr.get() };
-    if( !PyObject_Vectorcall( member->setattr_context, args, 3, 0 ) )
+    cppy::ptr ok( PyObject_Vectorcall( member->setattr_context, args, 3, 0 ) );
+    if( !ok )
         return -1;
     return 0;
 }
@@ -322,7 +324,8 @@ object_method_value_handler( Member* member, CAtom* atom, PyObject* value )
     cppy::ptr valueptr( member->full_validate( atom, Py_None, value ) );
     if( !valueptr )
         return -1;
-    if( !PyObject_CallMethodOneArg( pyobject_cast( atom ), member->setattr_context, valueptr.get() ) )
+    cppy::ptr ok( PyObject_CallMethodOneArg( pyobject_cast( atom ), member->setattr_context, valueptr.get() ) );
+    if ( !ok )
         return -1;
     return 0;
 }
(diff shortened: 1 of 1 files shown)


frmdstryr:
I'm also somewhat confused with the original code... doesn't it all need updated to use 

```
cppy::ptr ok(...)
if (!ok)
  return -1
```

??

MatthieuDartiailh (approved):
I agree that there is no reference stealing happening in the new version and that your usage of valueptr looks fine to me.

MatthieuDartiailh:
> I'm also somewhat confused with the original code... doesn't it all need updated to use 
> 
> ```
> cppy::ptr ok(...)
> if (!ok)
>   return -1
> ```
> 
> ??

Can you specify which line you are referring to ?

frmdstryr:
Like here https://github.com/nucleic/atom/blob/main/atom/src/setattrbehavior.cpp#L321

I think that is leaking a reference to the returned value?

MatthieuDartiailh:
I think you are correct. I need to look into stricter tests for references leaks !

sccolbert:
> Like here https://github.com/nucleic/atom/blob/main/atom/src/setattrbehavior.cpp#L321
> 
> I think that is leaking a reference to the returned value?

`setattr` is somewhat special in that it doesn't return a value. It either throws an exception (indicated by returning -1) or succeeds (indicated by returning 0).

sccolbert:
> Like here https://github.com/nucleic/atom/blob/main/atom/src/setattrbehavior.cpp#L321
> 
> I think that is leaking a reference to the returned value?

This is not leaking. `cppy::ptr` assumes you are supplying a new references to the constructor, and it decrefs that ownership on the destructor.

sccolbert:
Wait, I'm wrong again. Dammit I'm out of practice.

Yes, you are correct. The `PyObject_Call` returns a new reference. In the context of `setattr` this should be an exception or `Py_None`.

So, we do need to trap that in a `ptr` to be correct. Even though leaking refs to the `Py_None` singleton is not a huge deal.

MatthieuDartiailh:
But the return value from call is a PyObject*. In general it should be None which is immortal so leaking a reference does not matter but to be perfectly correct we should still decrease it, no ?

MatthieuDartiailh:
Then I made a not too bad mistake quite some time ago and we can now fix it. Good.

sccolbert (approved):

frmdstryr:
It might be less error prone make the handler's return a PyObject* and do one check  in Member::setattr?

sccolbert:
> It might be less error prone make the handler's return a PyObject* and do one check in Member::setattr?
> 
> Edit: Nevermind I guess that wouldn't always work.

While I wouldn't be opposed to that in theory, it would touch a lot of the other various `setattr` handlers, for little gain IMO. The `setattr` behavior really is unique within the Python C-api. My guess is that the core devs would change it if it wouldn't break the world.

frmdstryr:
It looks like full_validate is not called on the propery_handler value?

sccolbert:
> It looks like full_validate is not called on the propery_handler value?

IIRC that is correct. IIRC there wasn't enough room in the `Member` struct to store a validate handler *and* a setattr handler. So the assumption was that the setattr function would handle everything, since it *also* has to handle the storage of the actual value.

sccolbert:
> > It looks like full_validate is not called on the propery_handler value?
> 
> IIRC that is correct. IIRC there wasn't enough room in the `Member` struct to store a validate handler _and_ a setattr handler. So the assumption was that the setattr function would handle everything, since it _also_ has to handle the storage of the actual value.

Keep in mind that this library is well over 10 years old now and was heavily influenced by Enthought Traits. The `Property` member has always been a bit of a special case. Since it delegates the `get/set` behavior, it's very unclear who should own the actual value and the validation thereof.

MatthieuDartiailh:
To be honest the Property member and its behavior are weird to me. I will let @sccolbert comment.

sccolbert:
> To be honest the Property member and its behavior are weird to me. I will let @sccolbert comment.

Yeah. If I were to write the library again today, `Property` would not be a part of it.

MatthieuDartiailh:
Like you started the discussion on enaml maybe we could start a discussion on atom.

sccolbert:
> Like you started the discussion on enaml maybe we could start a discussion on atom.

Let's do it!!!!

PR #231, opened 2025-01-20
Update member.cpp to use fastcalls

frmdstryr:
In the `Member_set_x_mode` functions `cppy::replace` does an XINCREF even though it's already checked.

### DIFF
--- a/atom/src/member.cpp
+++ b/atom/src/member.cpp
@@ -105,14 +105,13 @@ Member_dealloc( Member* self )
 
 
 PyObject*
-Member_has_observers( Member* self, PyObject* args )
+Member_has_observers( Member* self, PyObject*const *args, Py_ssize_t n )
 {
-    const size_t n = PyTuple_GET_SIZE( args );
     if( n == 0 )
         return utils::py_bool( self->has_observers() );
     if( n > 1 )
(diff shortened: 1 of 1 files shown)


atom/src/member.cpp
```diff
@@ -457,17 +454,18 @@ Member_set_index( Member* self, PyObject* value )
 }
 
 
-template<typename T> bool
```
sccolbert, line 460: what's the reason for changing the return type here?

sccolbert (changes requested):
Most of this is a good change and is easy to reason about. I would prefer the changes to `parse_mode_and_context` to be a separate change with a corresponding reason.

atom/src/member.cpp
frmdstryr: I just thought it was cleaner since the size arg needed added anyways, but either works. I can revert it back if you prefer.

sccolbert (approved):

frmdstryr:
Thanks for reviewing all these PRs! Hopefully it makes everyone's code a little faster.

The only other thing I can think of possibly updating yet would be changing any of the tp_call's to use [vectorcallfunc](https://docs.python.org/3/c-api/call.html#the-vectorcall-protocol) but i'm not sure how to do that yet.

PR #232, opened 2025-01-20
Update catom to use fastcalls

frmdstryr:


frmdstryr:
I think CAtom_setstate needs reworked, idk what I was doing back then...

atom/src/catom.cpp
```diff
@@ -448,11 +445,8 @@ CAtom_getstate( CAtom* self )
 }
 
 PyObject*
-CAtom_setstate( CAtom* self, PyObject* args )
+CAtom_setstate( CAtom* self, PyObject* state )
 {
-    if( PyTuple_GET_SIZE( args ) != 1 )
-        return cppy::type_error( "__setstate__() takes exactly one argument" );
-    PyObject* state = PyTuple_GET_ITEM( args, 0 );
     cppy::ptr itemsptr = PyMapping_Items(state);
```
sccolbert, line 450: Do we need a `PyMapping_Check` before this?

That said, `PyMapping_Check` doesn't even check for `items()`.

Should we force it to be a `dict`?

Some of the below logic is questionable and may need some attention. I don't think that `__setstate__` should trigger observers, which means it shouldn't go thru the `__setattr__` protocol.

sccolbert (changes requested):
I agree with everything except the `__setstate__` method. I think that should be handled in a separate PR where we fix/agree on the behavior of that method.

frmdstryr:
Agree, I amended it to remove those changes.

PR #235, opened 2025-01-21
Correct handler range checks

frmdstryr:
My IDE was giving a warning because the if statements were always false. 

After looking at it more it looks like it was accidentally doing a size instead of array size check.

Since all the modes except for validate behavior have less than 15 handlers I made it fill out the array with the default handler and cast the mode to a 0xf so no if statement is needed.

### DIFF
--- a/atom/src/defaultvaluebehavior.cpp
+++ b/atom/src/defaultvaluebehavior.cpp
@@ -202,9 +202,12 @@ handlers[] = {
     call_object_object_name_handler,
     object_method_handler,
     object_method_name_handler,
-    member_method_object_handler
+    member_method_object_handler,
+    no_op_handler,
+    no_op_handler
 };
 
+static_assert( sizeof(handlers) / sizeof(handler) == 16, "Must be exactly 16 handlers" );
 
 }  // namespace
 
@@ -212,9 +215,7 @@ handlers[] = {
 PyObject*
 Member::default_value( CAtom* atom )
 {
-    if( get_default_value_mode() >= sizeof( handlers ) )
-        return no_op_handler( this, atom );  // LCOV_EXCL_LINE
-    return handlers[ get_default_value_mode() ]( this, atom );
+    return handlers[ get_default_value_mode() & 0xf ]( this, atom );
 }
 
 
--- a/atom/src/defaultvaluebehavior.cpp
+++ b/atom/src/defaultvaluebehavior.cpp
@@ -207,7 +207,7 @@ handlers[] = {
     no_op_handler
 };
 
-static_assert( sizeof(handlers) / sizeof(handler) == 16, "Must be exactly 16 handlers" );
+const auto mask = validate_handlers(handlers, DefaultValue::Mode::Last);
 
 }  // namespace
 
@@ -215,7 +215,7 @@ static_assert( sizeof(handlers) / sizeof(handler) == 16, "Must be exactly 16 han
 PyObject*
 Member::default_value( CAtom* atom )
 {
-    return handlers[ get_default_value_mode() & 0xf ]( this, atom );
+    return handlers[ get_default_value_mode() & mask ]( this, atom );
 }
 
 
--- a/atom/src/validatebehavior.cpp
+++ b/atom/src/validatebehavior.cpp
@@ -1007,7 +1007,7 @@ handlers[] = {
 PyObject*
 Member::validate( CAtom* atom, PyObject* oldvalue, PyObject* newvalue )
 {
-    if( get_validate_mode() >= sizeof( handlers ) )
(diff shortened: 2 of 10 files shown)


frmdstryr:
I think the extra_modes field could be deleted from member if everything other than validate mode was collapsed to a size of 0xf?

MatthieuDartiailh (approved):

frmdstryr:
I guess there's no need to add so many empty entries, let me reduce some of them.

frmdstryr:
I did my static assert's wrong.. it should be 16.

sccolbert:
Let me know when you are done making changes, and I will review.

frmdstryr:
Ok, I think it's good now.

frmdstryr:
Since the Validate enum size is right at the limit of 5 bits do you think it should do a check on the handler size (in case someone tries to add something later)?

atom/src/validatebehavior.cpp
```diff
@@ -1007,7 +1007,7 @@ handlers[] = {
 PyObject*
 Member::validate( CAtom* atom, PyObject* oldvalue, PyObject* newvalue )
 {
-    if( get_validate_mode() >= sizeof( handlers ) )
+    if( get_validate_mode() >= sizeof( handlers ) / sizeof( handler ) )
```
sccolbert, line 1010: This looks different that the rest?

sccolbert:
Are we making this change just to satisfy an IDE linter?

sccolbert:
I think all we really need to static assert is that the largest `**Mode` enum value is `< sizeof(**handlers)`.

sccolbert:
That way, things could still grow more handlers in the future if need-be.

frmdstryr:
Agree, let me update it.

sccolbert:
Maybe we convert these C-arrays into C++ arrays:
https://en.cppreference.com/w/cpp/container/array

Then we don't have to do the division to compute the length.

frmdstryr:
Good idea, do you want to take over or should I just update this?

sccolbert:
I think you are well equipped to do it. I'm just a gray-hair giving advice nowadays.

atom/src/defaultvaluebehavior.cpp
```diff
@@ -212,8 +211,6 @@ handlers[] = {
 PyObject*
 Member::default_value( CAtom* atom )
 {
-    if( get_default_value_mode() >= sizeof( handlers ) )
-        return no_op_handler( this, atom );  // LCOV_EXCL_LINE
     return handlers[ get_default_value_mode() ]( this, atom );
```
MatthieuDartiailh, line 214: Here you basically eliminated bound checks. It may be better to use std::array::at to benefit from the stdlib bound checks.

atom/src/defaultvaluebehavior.cpp
```diff
@@ -187,8 +187,7 @@ typedef PyObject*
 ( *handler )( Member* member, CAtom* atom );
 
 
-static handler
-handlers[] = {
+static const std::array<handler, DefaultValue::Mode::Last> handlers = {
```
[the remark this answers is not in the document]
frmdstryr: I think it just implicitly casts the enum to size_t?

atom/src/defaultvaluebehavior.cpp
frmdstryr: The check is done when it parses the enum value, but I agree it's still possible to go OOB if the mode bits somehow get misparsed or that check is bypassed.

My goal of originally looking at this was to try and delete if's... I'd rather go back to re-adding no_op_handlers than add another if.

frmdstryr: `std::array<handler, 8>` doesn't give a compile error if you give less than 8 entries so I'd rather just go back to https://github.com/nucleic/atom/pull/235/commits/8b32c1b90ee0f47b41662cf7e679a2ce32cbb271 with maybe an additional static_assert check on the sentinel?

MatthieuDartiailh: Can we statically assert that we do not have more enum variants than we have handlers?  if we can do that we should be good.

MatthieuDartiailh (approved):
@sccolbert do you want to have another look ?

sccolbert:
Yes, I'd like a stab at it. Please don't merge this yet. 

I will make another PR with a differing implementation of these same ideas.

MatthieuDartiailh:
@sccolbert are you still planning to make an alternate version ?

sccolbert:
I think there is a cleaner way of doing this, but I dont have the time
right now. If it works and solves the current problem, I'm okay with it. We
can always improve it later.

PR #236, opened 2025-01-21
Use packed struct with bitfields to shrink all member modes down to 32-bits

frmdstryr:
All of the modes can fit into 32 bits. This shrinks the member struct down to 128 bytes (from 144) with an alignment of 8.

The compiler also nicely does all the bit fiddling.

### DIFF
--- a/atom/src/behaviors.h
+++ b/atom/src/behaviors.h
@@ -14,7 +14,7 @@ namespace atom
 namespace GetAttr
 {
 
-enum Mode
+enum Mode: uint8_t
 {
     NoOp,
     Slot,
@@ -37,7 +37,7 @@ enum Mode
 namespace PostGetAttr
 {
 
-enum Mode
+enum Mode: uint8_t
 {
     NoOp,
     Delegate,
@@ -53,7 +53,7 @@ enum Mode
 namespace SetAttr
 {
 
-enum Mode
+enum Mode: uint8_t
 {
     NoOp,
     Slot,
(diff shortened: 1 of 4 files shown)


MatthieuDartiailh (commented):
I personally love getting rid off the manual masking.

Out of curiosity what is the advantage to move modes and index at the end of the Member struct ?

frmdstryr:
My understanding is compilers will insert padding bytes depending on the size of fields and alignment. So if you have a struct like

```
struct Foo {
   uint64_t a;
   uint32_t b; 
   uint64_t c;
};
``` 

Unless it is packed the compiler might/will insert 4 padding bytes after b so all fields are 8 byte aligned. By sorting fields by their size it minimizes any padding.

MatthieuDartiailh (dismissed):

frmdstryr:
I just realized it works using

```cpp
PACK(struct MemberModes
{
    GetAttr::Mode getattr: 4;
    PostGetAttr::Mode post_getattr: 3;
    SetAttr::Mode setattr: 4;
    PostSetAttr::Mode post_setattr: 3;
    DefaultValue::Mode default_value: 4;
    Validate::Mode validate: 5;
    PostValidate::Mode post_validate: 3;
    DelAttr::Mode delattr: 3;
    GetState::Mode getstate: 3;
});
```

So the get/set methods could just all be dropped.

sccolbert:
Neat trick! I'll remember this next time I need bitfields.

PR #240, opened 2025-02-01
Draft: Refactor catom to reduce size of unobserved atoms

frmdstryr:
@MatthieuDartiailh's discussion on "un-observable atoms" got me thinking.. Currently there is 4 bytes of wasted space in the CAtom struct due to padding (on 64bit systems) also even if atoms have no observers the ObserverPool* still takes 8 bytes of space. 

This PR proposes a potential new structure that converts the `ObserverPool* observers` field into a `uint32_t pool_index` and collapses that into a packed struct with the bitfield. The new catom struct has no padding and is now 32 bytes (down from 40 bytes).

Since the ObserverPool* is gone it uses a new "ObserverPoolManager" and a new bitfield flag "has_observers" that manages the mapping of the pool_index to ObserverPool for each atom. 

Since it reuses ObserverPools (instead of delete & new) it is quite a bit faster add adding/removing observers however the observer pool manager (currently) doesn't ever shrink in size. 

So atoms with no observers save 8 bytes, but atoms with observers cost 4 more bytes if my math is right (since pointers are just moved to the ObserverPoolManager and the extra free slot takes 4 bytes per pool). A better pool manager could eliminate the free slots vector to make it 0 cost if there's any ideas?

Just throwing this out there for discussion, there's probably better ways to do it.

### DIFF
--- a/atom/src/catom.cpp
+++ b/atom/src/catom.cpp
@@ -102,9 +102,10 @@ CAtom_clear( CAtom* self )
     {
         Py_CLEAR( self->slots[ i ] );
     }
-    if( self->observers )
+    if( self->meta.has_observers )
     {
-        self->observers->py_clear();
+        ObserverPoolManager::get()->release_pool(self->meta.pool_index);
+        self->meta.has_observers = false;
     }
 }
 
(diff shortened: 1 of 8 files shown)


frmdstryr:
Well idk why it says the `sizeof(ObserverPool*)` is 24 bytes, it's 8 on my computer...

frmdstryr:
Interesting... will have to do some more reading. From a quick scan I can't see how subinterpreters improve anything (given their examples shows pickling objects between them).

frmdstryr:
I realize this PR is likely not going anywhere but tried changing the ObserverPoolManager 

```
std::vector<ObserverPool> m_pools;
```

Which should delete 8 bytes per observed atom... but doing so causes segfaults when running  the example from https://github.com/nucleic/atom/pull/239#issuecomment-2613099374 for reasons I can't explain. Yet it passes all atom tests and enaml tests...

This makes me think something funny is going on when it catom clears the pool and there are active modification guards.

frmdstryr:
This will just cause more issues going forward due to the upcoming changes mentioned by Matthieu.

PR #242, opened 2025-02-01
Use emplace instead of insert/push in ObserverPool::add

frmdstryr:
This avoids an extra xincref/xdecref on the cppy::ptr because insert & push_back do a copy. 

See https://godbolt.org/z/zGvTzGahn

The `*obs_free = Observer()` does the same thing but I'm not sure how to cleany fix that (other than manually setting the fields).


sccolbert:
For the last part, I think you can do:
```cpp
obs_free->m_change_types = change_types;
obs_free->m_observer = observer;
```

PR #243, opened 2025-02-04
Replace CAtom's bitfield with packed struct

frmdstryr:
Reduce manual bit fiddling...


MatthieuDartiailh (approved):

frmdstryr:
I made one more micro optimization.... In CAtom_new it currently calls `atom->set_notifications_enabled( true );`, instead of doing a second write (since it rewrites when setting slots), I inverted the meaning of the notifications bit so it is on by default.  

I haven't looked if the compiler is smart enough to optimize it into a single write or not but this way it should only ever be one write.

PR #248, opened 2025-09-29
Fix type hint

AlexanderS:
Sphinx does not understand forwared references without quotes. So autodoc will fail if the quotes are missing.

You can see, that the modules is missing in the official documentation: https://atom.readthedocs.io/en/latest/api/atom.datastructures.html

You can see the error while building the documentation f.e. here: https://github.com/nucleic/atom/actions/runs/17932608666/job/50992636914


AlexanderS:
The CI error seems unrelated to this fix.

MatthieuDartiailh:
Thanks for the fix

PR #263, opened 2026-04-29
Update sortedmap

frmdstryr:
This slightly reworks the constructor to add support for tp_vectorcall (~30% faster construction of empty sortedmap when available) and addresses the recent code review items.

Edit: Also adds a special case for doing a sortedmap copy in the constructor.

### DIFF
--- a/atom/src/sortedmap.cpp
+++ b/atom/src/sortedmap.cpp
@@ -131,6 +131,53 @@ struct SortedMap
 
 	static bool Ready();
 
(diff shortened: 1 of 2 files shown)


frmdstryr:
I was looking at this more. Doesn't enaml use sortedmap with empty objects for as scope keys by default? This means nearly every comparison is raising an error, squashing it, then falling back to ptr comparisons...

MatthieuDartiailh:
I do not remember and had no time to check. I am wondering if we should not move this specialized container directly to enaml and optimize it as needed.

PR #264, opened 2026-04-30
Add test for cached property error case

frmdstryr:
The code review claimed there was an issue with the getattr cached_property_handler but I see no problems. This adds test coverage for the error case.

### DIFF
--- a/tests/test_property.py
+++ b/tests/test_property.py
@@ -178,7 +178,8 @@ def test_cached_property_error():
     with pytest.raises(ValueError) as excinfo:
         o.x  # This puts null in the slot
     assert "get_x failed" in excinfo.exconly()
-    assert o.x == v
+    assert o.x == 2
+    assert o.x == 2
 
 
 def test_observed_property():
--- a/tests/test_property.py
+++ b/tests/test_property.py
@@ -160,6 +160,27 @@ def test_enforce_read_only_cached_property():
         p.setter(set)
 
 
+def test_cached_property_error():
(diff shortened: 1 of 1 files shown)


tests/test_property.py
```diff
@@ -160,6 +160,21 @@ def set(self, value):
...
+    class Obj(Atom):
+        def _get_x(self):
+            raise ValueError("get_x failed!")
+
+        x = Property(_get_x, cached=True)
+
+    o = Obj()
+    with pytest.raises(ValueError) as excinfo:
+        o.x
+    assert "get_x failed" in excinfo.exconly()
```
MatthieuDartiailh, line 180: It would be worth accessing the value a second to exercise the caching behavior in presence of error.

frmdstryr: Good point.

tests/test_property.py
```diff
@@ -160,6 +160,27 @@ def set(self, value):
...
+                raise ValueError("get_x failed!")
+            return v
+
+        x = Property(_get_x, cached=True)
+
+    o = Obj()
+    with pytest.raises(ValueError) as excinfo:
+        o.x  # This puts null in the slot
+    assert "get_x failed" in excinfo.exconly()
+    assert o.x == v
```
MatthieuDartiailh:
```suggestion
    assert o.x == 2
    assert o.x == 2
```
Check the cache does work on first successful call

MatthieuDartiailh (approved):

PR #266, opened 2026-04-30
Update atomdict

frmdstryr:
This validates the key before use in AtomDict_setdefault and DefaultAtomDict_missing to avoid the unlikely yet possible case where the key is coerced which would cause the second lookup to return null. 

It also corrects the bases argument as described in the recent code review, checks for unicode decode errors in DefaultDict_repr, cleans up some outdated version guards, and changes DefaultAtomDict_missing to METH_O.

### DIFF
--- a/atom/src/atomdict.cpp
+++ b/atom/src/atomdict.cpp
@@ -103,11 +103,7 @@ int AtomDict_traverse( AtomDict* self, visitproc visit, void* arg )
 {
 	Py_VISIT( self->m_key_validator );
 	Py_VISIT( self->m_value_validator );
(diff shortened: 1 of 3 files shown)


MatthieuDartiailh:
Out of curiosity do you have a use case for this coercion of the key before access ? I agree it is correct but it feels unfortunate. I wonder if we should validate only if the key is not found to avoid the performance hit.

frmdstryr:
I've never used it no.

frmdstryr:
It could do an unvalidated key lookup first, if it fails, lookup then do what is it does in this pr? Or do you have another idea?

PR #270, opened 2026-05-08
Update atomlist

frmdstryr:
This fixes a leak in validate_sequence, updates insert & pop to use vectorcalls, and cleans up some outdated version guards.

Since full_validate already returns a new reference it does not need an extra incref, it just gets stolen by the list. PySequence_List returns a populated copy of the items. PyList_SET_ITEM is intended to be used with PyList_New so it does not discard the old item meaning the old item was leaked.

### DIFF
--- a/atom/src/atomlist.cpp
+++ b/atom/src/atomlist.cpp
@@ -97,11 +97,7 @@ ListSubtype_New( PyTypeObject* subtype, Py_ssize_t size )
             return PyErr_NoMemory();  // LCOV_EXCL_LINE (memory error)
         memset( op->ob_item, 0, nbytes );
     }
-#if PY_VERSION_HEX >= 0x03090000
     Py_SET_SIZE( op, size );
-#else
-    Py_SIZE( op ) = size;
-#endif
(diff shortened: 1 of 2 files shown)


atom/src/atomlist.cpp
```diff
@@ -248,7 +246,8 @@ class AtomListHandler
                     PyObject* val = vd->full_validate( atm, Py_None, b );
                     if( !val )
                         return 0;
-                    PyList_SET_ITEM( templist.get(), i, cppy::incref( val ) );
+                    PyList_SET_ITEM( templist.get(), i, val );
+                    Py_DECREF( b ); // Release old reference
```
MatthieuDartiailh: PyList_GET_ITEM return a borrowed reference but PyList_SET_ITEM does not decrement the ref count of the replaced object. So this is correct but really deserves a comment.

frmdstryr: Agreed. Added a comment there to clarify.

frmdstryr:
Thanks!

PR #272, opened 2026-05-29
Add Bool(strict=False) to to use truthyness of object

frmdstryr:
This adds a new bool_promote validation mode that uses the value's truthyness to determine the saved value. 

Since this pushes the number of validators over 32 the enum needs another bit. To avoid adding extra padding (and make room for any potential future validators), the index was moved into the packed struct and changed to uint16_t (since catom only supports u16 anyways).

Using a negative index is now an error. This is potentially breaking but I think also a user error so that's ok in my opinion.

### DIFF
--- a/atom/scalars.py
+++ b/atom/scalars.py
@@ -85,7 +85,7 @@ class Bool(Value):
     """A value of type `bool`.
 
     By default, bools are strictly typed.  Pass strict=False to the
-    constructor to use the result truthyness of the value (eg bool(value)).
+    constructor to use the truthyness of the value (eg bool(value)).
 
     """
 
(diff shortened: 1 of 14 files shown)


MatthieuDartiailh:
I am not a fan of adding another coercion like behavior. But I can understand the need. Could you add changelog entry ?

sccolbert:
Apparently GH is now blocking my email replies, so here I am... :sigh:

> Can't you already do this with a `Coerced` atom? That's why I made that Atom, for the rare case that you need to coerce something to a different type...

> Pretty sure `Coerced(bool)` is all you need for this...

sccolbert:
If it's a perf issue, I'd need to see some evidence for a real use case where it makes a real difference...

frmdstryr:
Yes Coerced(bool) can be used instead, but there's already strict arguments for the Int, Float, Str, and Bytes members so this is mainly for consistency.  If it's unwanted then this can be closed, it's not a big deal.

PR #273, opened 2026-09-09
Wrap use of PyWeakref_GET_OBJECT if we need it

s-t-e-v-e-n-k:
Python 3.15 removes PyWeakref_GET_OBJECT, and the path forward is by using PyWeakref_GetRef, which was only added in Python 3.13. Add a wrapper for it if required, and switch to using PyWeakref_GetRef everywhere that called PyWeakref_GET_OBJECT.


MatthieuDartiailh:
Thanks for your contribution. Could you also add 3.15 to the test and release matrix ?

s-t-e-v-e-n-k:
> Thanks for your contribution. Could you also add 3.15 to the test and release matrix ?

I have not added it to the release matrix for now, since 3.15 final isn't out quite yet.

MatthieuDartiailh:
The typing error is unrelated so I will merge thanks

